PR #8, opened 2018-06-22
MYFACES-4239: Multiple performance improvements

pnicolucci:


tandraschko:
@pnicolucci i'm not sure about this changes out.write vs sb.append...

Why do you switchted from out.write to a new StringBuilder?
You know, this are always new object instances... And we have something like a SharedString

pnicolucci:
@tandraschko  , we've been running some performance tests and can see a .5 - 1% increase in our testing in performance with an application that is only about 15% JSF , with a larger application and more JSF within it , the gain could be even greater.  Basically we want to avoid the writer.write path , it is quite a long path to call over and over again if it isn't necessary.

tandraschko:
ok, i see
i'm ok for it if you use the SharedStringBuilder

tandraschko:
JFYI: we avoided new objects so much in the past that we even switched from for-each to for loops, so the SharedStringBuilder is IMO a must-have for this change

tandraschko:
All other changes looks fine of course.

pnicolucci:
I'll refactor a bit to use the SharedStringBuilder. Thanks for the information / review!

tandraschko:
I also thought a bit about it, i think it would be better if we discuss it first on the mailing list.
And details about it would be great. You know, we already have a buffer.

So could you maybe split the sb/out commits from the other improvements?

pnicolucci:
I'll work to remove the sb changes from this PR and will open a new issue / PR for the changes and get a discussion started on the mailing list.

pnicolucci:
@tandraschko  ok with the current changes in this PR?

tandraschko:
Yep! :)

PR #10, opened 2018-07-11
MYFACES-4244:Use StringBuilder rather than calling write multiple times

pnicolucci:
Creating this PR for discussion on the dev list for MYFACES-4244: https://issues.apache.org/jira/browse/MYFACES-4244


pnicolucci:
After additional investigation we've determined this change is not necessary so I'm closing this Pull request.

PR #11, opened 2018-07-12
MYFACES-4247: reduce number of char[] objects and CommonPropertyUtils…

pnicolucci:
… calls


pnicolucci:
@tandraschko  can you review these changes and let me know if you have any concerns? If not I'll merge this PR.

tandraschko:
+1 for the CommonProperties stuff

+0 for the changes in write
could you explain them before?

tandraschko:
if it works fine, i'm ok with it
it would be just better if we would have a unittest for it?

pnicolucci:
I'll add a couple unit tests and ensure the behavior is the same before an after the changes just to be extra safe.

PR #18, opened 2018-08-26
MYFACES-4082 Fix 'binding' when referencing composite components

luigiJsonar:
Code that 'sets' the value of the binding attribute for composite components. Inspired on:
org.apache.myfaces.view.facelets.tag.jsf.ComponentTagHandlerDelegate#createComponent

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/view/facelets/tag/composite/CompositeComponentResourceTagHandler.java
+++ b/impl/src/main/java/org/apache/myfaces/view/facelets/tag/composite/CompositeComponentResourceTagHandler.java
@@ -42,6 +42,7 @@ import javax.faces.component.UIPanel;
 import javax.faces.component.UniqueIdVendor;
 import javax.faces.component.ValueHolder;
 import javax.faces.context.FacesContext;
+import javax.faces.event.PhaseId;
 import javax.faces.view.AttachedObjectHandler;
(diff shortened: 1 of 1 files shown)


tandraschko:
@luigiJsonar 

the ComponentTagHandlerDelegate contains the following, before setting the value:
if (PhaseId.RESTORE_VIEW.equals(faces.getCurrentPhaseId()))

Could you please if it makes sense to add this if?

tandraschko:
@luigiJsonar 

Could you please try this:

https://gist.github.com/tandraschko/2f8e15bcbf8affc63df26b25f9259584

If it works fine, i will apply it

luigiJsonar:
Applied the suggested changes and it works fine.

luigiJsonar:
I'd be more than happy to apply any other suggestions, open more PRs for previous versions, test it... I don't understand why it had to be your commit, not mine. Can you point me to any documentation on how to contribute to this repo? I need that because I might be missing something but IMHO I tried to contribute and you just got my code and didn't even mention me on it.

tandraschko:
Sorry, please don't take it personally. I didn't see that you applied my changes, so i commited my version, which was inspired by your version and the original code. 

You can just contribute by opening issues, trigger discussions on the dev mailinst list and provide patches/Pull Requests.

PR #34, opened 2018-12-05
MYFACES-4273 close() JarFile in Classpath

wtlucy:
see https://issues.apache.org/jira/browse/MYFACES-4273


tandraschko:
basically +1
however a try/resource statement would be a bit nicer probably :)

wtlucy:
@tandraschko how's this?  I'll still need to use the try finally method for the 2.0 and 2.1 branches

tandraschko:
looks good if it still works the same :)

wtlucy:
Yeah, it's the same logic.  I'll put this fix in 2.2, 2.3, and 3.0.

PR #46, opened 2019-03-12
MYFACES-4282 skip clearing view map during view metadata creation

wtlucy:
For https://issues.apache.org/jira/browse/MYFACES-4282


tandraschko:
@wtlucy would you like to add 2 unittests for both cases?

So we could release the next version in May - the last release was in Januar.

wtlucy:
@tandraschko sure, I'll plan to add some tests for this next week.

wtlucy:
@tandraschko I've added some really basic tests here.  Any comments?

tandraschko:
+1
thanks
please apply it :)

tandraschko:
@wtlucy thanks! please don't forget 3.x

wtlucy:
@tandraschko done :) - https://github.com/apache/myfaces/pull/50

PR #58, opened 2019-06-13
fix: do not treat spaces as invalid characters

cnsgithub:
This might hurt performance.


tandraschko:
Can you please also adjust the unittest then?

cnsgithub:
@tandraschko Could not do that within the same PR via github site, sorry about that. Additional PR: https://github.com/apache/myfaces/pull/59

PR #59, opened 2019-06-13
fix: adjust unit test to new character ranges

cnsgithub:


tandraschko:
is "U+D7FF" really not invalid anymore?

cnsgithub:
Yep, see https://www.w3.org/TR/REC-xml/#charsets:

image

tandraschko:
@cnsgithub thanks! If you would like to see it in 2.3.x, please create a JIRA issue and a PR for this branch.

PR #69, opened 2019-09-19
MYFACES-4300 upgrade commons-beanutils to 1.9.4 - 2.1.x

wtlucy:
myfaces is not vulnerable, but this library should be updated as a best practice


tandraschko:
+1 - Just apply it in all branches

wtlucy:
@tandraschko done

PR #77, opened 2020-01-09
Myfaces 4311 Unit Tests

volosied:
Added two unit tests to ensure that the generic converter & validator work as expected.

And let me know if there are any issues. Particularly, I'm not sure if I should implement testValidator another way.  I kept it simple with the try catch, but I can use some JUnit feature to expect a FacesException if recommended. 

Tested issue:
https://issues.apache.org/jira/browse/MYFACES-4311
https://github.com/apache/myfaces/pull/75


wtlucy:
This looks good @volosied, thanks for circling back to add these tests.

PR #79, opened 2020-01-15
MYFACES-3822 remove unused params from JSF 2.2.x

volosied:
A number of unused params were removed, but they have been remained in MyFacesConfig

https://issues.apache.org/jira/browse/MYFACES-3822

For backwards compatibility (based on comments), the following were left alone 
org.apache.myfaces.HANDLE_STATE_CACHING_MECHANICS org.apache.myfaces.SAVE_STATE_WITH_VISIT_TREE_ON_PSS
org.apache.myfaces.ERROR_HANDLER 
org.apache.myfaces.CACHE_OLD_VIEWS_IN_SESSION_MODE

### DIFF
--- a/shared/src/main/java/org/apache/myfaces/shared/config/MyfacesConfig.java
+++ b/shared/src/main/java/org/apache/myfaces/shared/config/MyfacesConfig.java
@@ -54,29 +54,6 @@ public class MyfacesConfig
     public static final String  INIT_PARAM_RESOURCE_VIRTUAL_PATH_DEFAULT = "/faces/myFacesExtensionResource";
 
     /**
-     * If true, rendered HTML code will be formatted, so that it is "human readable".
-     * i.e. additional line separators and whitespace will be written, that do not
-     * influence the HTML code. Default: "true"
-     */
-    @JSFWebConfigParam(defaultValue="true", expectedValues="true, false, on, off, yes, no",since="1.1",
-            ignoreUpperLowerCase=true, group="render")
-    private static final String  INIT_PARAM_PRETTY_HTML = "org.apache.myfaces.PRETTY_HTML";
-    private static final boolean INIT_PARAM_PRETTY_HTML_DEFAULT = true;
-
-    /**
-     * This parameter tells MyFaces if javascript code should be allowed in the rendered HTML output.
-     * If javascript is allowed, command_link anchors will have javascript code 
-     * that submits the corresponding form.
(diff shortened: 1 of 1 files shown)


volosied:
Some of the tests will also need to be changed.

tandraschko:
Hmpf, i general i don't have a problem to do such cleanups, however i personally see 2.2 and everything older in maintanence mode
@pnicolucci WDYT?

volosied:
I see, but I'll go ahead and continue making similar changes to 2.3 and master! Thanks!

volosied:
Been a bit busy, but I will get the other branches done soon!

pnicolucci:
@volosied  I agree with @tandraschko  here. It makes more sense to do the clean up in the  2.3 and master branches. Can you explain what you mean by your comment:

For backwards compatibility (based on comments), the following were left alone
org.apache.myfaces.HANDLE_STATE_CACHING_MECHANICS org.apache.myfaces.SAVE_STATE_WITH_VISIT_TREE_ON_PSS
org.apache.myfaces.ERROR_HANDLER
org.apache.myfaces.CACHE_OLD_VIEWS_IN_SESSION_MODE

I didn't actually see those in the MyFacesConfig, did I miss them? 

Also the commit history would be a bit easier to follow here if we just used one commit to remove these params rather than one for each removal, just my opinion.

volosied:
@pnicolucci In that case, I can close this PR then without merging it in. 

As for your question, I saw comments saying those 4 parameters were deprecated, but after taking a closer look, I believe their functionality was removed.  (Since I don't see mentions of the them anywhere else in the code.)  They aren't  in the MyFacesConfig, but in the following files: 

org.apache.myfaces.HANDLE_STATE_CACHING_MECHANICS 
   impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java

org.apache.myfaces.SAVE_STATE_WITH_VISIT_TREE_ON_PSS
   impl/src/main/java/org/apache/myfaces/view/facelets
/DefaultFaceletsStateManagementStrategy.java

org.apache.myfaces.ERROR_HANDLER
   api/src/main/java/javax/faces/webapp/PreJsf2ExceptionHandlerFactory.java

org.apache.myfaces.CACHE_OLD_VIEWS_IN_SESSION_MODE
   impl/src/main/java/org/apache/myfaces/application/viewstate/ServerSideStateCacheImpl.java

tandraschko:
yep, lets close it

PR #80, opened 2020-01-22
MYFACES-4318 check for incorrect object instance during restore

wtlucy:
Proposed fix for https://issues.apache.org/jira/browse/MYFACES-4318


tandraschko:
lets close this for now in favor of docs / leo suggestion

PR #84, opened 2020-02-26
MYFACES-4320 Startup ConcurrentModificationException 2.3.x

wtlucy:
For https://issues.apache.org/jira/browse/MYFACES-4320


tandraschko:
+1

PR #89, opened 2020-03-08
Support for PrimeFaces 8 in Quarkus native mode

rmpestano:


rmpestano:
Even though we are using PrimeFaces 8 on the runtime of myfaces extension, users still can use PF 7. For example I've just enabled native mode on [quarkus-omnifaces](https://github.com/rmpestano/quarkus-omnifaces) extension and it's sample app uses latest myfaces and PrimeFaces 7.

PR #90, opened 2020-03-27
Incorrect Last-Modified header for resources

astappiev:
We conduct an issue that our resources have incorrect header: `Last-Modified: Thu, 01 Jan 1970 00:00:00 GMT`.

After investigating the issue, I found that we deploy using Tomcat with a version tag and a path to resources contains special characters (`#`).
And this method can not read the resources, because the URL (file path) contains URL Encoded characters (for OS this is not a valid path) like this: `/some-path/tomcat/webapps/v3%23%233.2.1/resources/layout/js/layout.js`.

I'm not sure if this PR is the best solution. But it solves this issue for us.
Now I created a custom implementation of ResourceHalder and Resource (just a copy of myfaces impls) to update the method and fix this issue. It would be good if you can release a fix, or give a hint on how to patch it easier.

### DIFF
--- a/shared/src/main/java/org/apache/myfaces/shared/resource/ResourceLoaderUtils.java
+++ b/shared/src/main/java/org/apache/myfaces/shared/resource/ResourceLoaderUtils.java
@@ -18,7 +18,6 @@
  */
 package org.apache.myfaces.shared.resource;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.JarURLConnection;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


tandraschko:
Could you please create a JIRA issue first? (thats how someone should start before creating a PR :D, thanks!)

astappiev:
https://issues.apache.org/jira/browse/MYFACES-4326

tandraschko:
@astappiev Thanks!

Could you please try to remove the "file".equals-statement and always call 
`getResourceLastModified(url.openConnection());` ?
Does this work?

astappiev:
Yes, all values as expected.

tandraschko:
ok
could you please modify the PR?
And it would be great if you can also add the changes to 3.0.x branch and master branch

tandraschko:
looks good, thanks!
can you also provide it for the other branches i mentioned? so i can merge it and close the issue

PR #92, opened 2020-04-04
Add Ukrainian translation for Faces Messages

astappiev:


tandraschko:
Same here, please add a JIRA first 
and also please create PRs for at least 2.3.x, 3.0.x and master branch

astappiev:
I found a few more messages bundles:
- shared\src\test\resources\javax\faces\Messages.properties
- shared-public\src\test\resources\javax\faces\Messages.properties
- api\src\main\javascript\META-INF\resources\myfaces\_impl\i18n\Messages.js

Does it make sense to translate any of them? In particular, the first two are in the `test` folder.

werpu:
Hi I only can speak for the messages.js, no need to translate those, we will drop the multi language support in the next gen of the jsf.js, from then onwards only english messages will be shown coming out of the framework (not that the end user ever sees those). But that is just the javascript part, cannot comment on the rest.

tandraschko:
src/test can just be ignored, it's really only for the unittests

tandraschko:
any news? :)

astappiev:
Hi,

I added all other messages files that we updated (used in our project);
- `Messages.properties`: removed few trailing spaces
- `Messages_en.properties`: same as above
- `Messages_de.properties`: fixed character encodings, spaces
- `Messages_es.properties`: encodings, updated order (to compare with base more easily), added missing variables (like `{2}: ` prefixes)
- `Messages_it.properties`: like above
- `Messages_pt.properties`: copied from `Messages_pt_BR.properties`, updated like above

Will it be useful to merge? If not, I will revert and only finish Ukrainian translation.

tandraschko:
Yep of course
I really wonder why we have wrong encoding and nobody mentioned it over the years

astappiev:
What to do with Portuguese and Brazilian Portuguese?

We have a single language for both, and therefore we renamed `Messages_pt_BR.properties` to `Messages_pt.properties` and updated the last one. 

In this repository, I created a new file, but does it makes sense to keep two files identical (right now Messages_pt.properties is slightly better, but as we didn't change any sentences it can be backported)? Should I rename `Messages_pt_BR.properties` to `Messages_pt.properties` in the PR?

astappiev:
https://issues.apache.org/jira/browse/MYFACES-4331

astappiev:
Where does `org.apache.myfaces.renderkit.html.HtmlMessagesRenderer.IN_FIELD` message used? There is only one word and it is untranslated in many languages, should I keep it as is?

```
org.apache.myfaces.renderkit.html.HtmlMessagesRenderer.IN_FIELD = \u0020in {0}
```
* `\u0020` is a space character

tandraschko:
i think it's unused
we should remove it in master at least

tandraschko:
about pt pt_BR:
i think we should leave it as it is in older branches and rename it in master

astappiev:
Created additional merge requests for `master` and `3.0.x` branches.
- In master, I renamed `pt_BR` to `pt`. In other branches added additionally.
- I didn't remove any fields anywhere

PR #96, opened 2020-04-18
Update messages translations, add Ukrainian, rename Portuguese

astappiev:


tandraschko:
You also added a new file Messages_uk.properties
it contains only escaped chars, is this correct?

astappiev:
> You added the new file Messages_uk.properties - it contains only escaped chars, I assume this is correct? (never checked the Ukraine language, so no idea :))

Unfortunately, we still use the Cyrillic alphabet and all characters our of ASCII range. Similar to Chinese, Arabic...

> Another question is e.g.:
javax.faces.validator.RegexValidator.NOT_MATCHED = {1}: Validation Error: Value not according to pattern ''{0}''
> is it correct hat {0} is surrounded by double '?

I don't know, they wrapped into double '' in the source file (messages.properties).

tandraschko:
ok i see, will merge it for now
thank you for your great contribution!

tandraschko:
@astappiev could you please check my latest commit? I removed the IN_FIELD message via my IDE and it only showed those removal changes. However github shows some encoding changes....

astappiev:
All fine, your IDE replaced upper case letters with lower case for native-encoded characters.
\+ Fixed encoding for all other languages.

PR #98, opened 2020-04-21
Rename myfaces-quarkus-runtime to myfaces-quarkus

gastaldi:


tandraschko:
Thanks for the PR!
everything else is now aligned with the quarkus naming conventions? (like directory structure?)

gastaldi:
The directories look ok, but you need to add a `quarkus-extension.yaml` in the runtime module: https://github.com/quarkusio/quarkus/blob/master/extensions/resteasy-qute/runtime/src/main/resources/META-INF/quarkus-extension.yaml

tandraschko:
Thanks, i will add one

tandraschko:
@gastaldi i wonder how it should work

should a user only add the quarkus-myfaces dependency or also the quarkus-myfaces-deployment?
in your tutorials, you only add the runtime module (https://quarkus.io/guides/qute)
how is the deployment module important in "showcase" then?

gastaldi:
The deployment module is transparently (or magically? :) )  resolved by the quarkus plugin, in the showcase you must ONLY depend on the runtime module.

tandraschko:
Will this also work in our case for external quarkus extensions?

tandraschko:
@gastaldi could you do me favor?
our showcase doesn't start anymore and i'm unsure if it's related to your artifact-id change or related to quarkus 1.3.2

also removing the deployment-dependency in the showcase doesnt work for me

tandraschko:
ok NVM - recompiled everything and works now

but removing the deployment-module from showcase doesnt work
so i think the quarkus plugin doesn't resolve external extensions

gastaldi:
@aloubyansky could you confirm that this is an issue?

aloubyansky:
I guess it's because the quarkus maven plugin isn't properly configured for the runtime artifact. @tandraschko have you followed https://quarkus.io/guides/writing-extensions#maven-setup ? You should see `META-INF/quarkus-extension.properties` in the runtime JAR. Users should never add the deployment deps to their apps.

tandraschko:
ahh i see
works now, thanks :)

PR #99, opened 2020-04-26
Disable method handles in quarkus native mode

rmpestano:


tandraschko:
thanks for the PR!
I just wonder why we need common-lang now. Did quarkus add this dependency?

rmpestano:
About the commons lang, probably it was being brought by another dependency transitively before because after rebasing my branch with master I started getting compilation error,

rmpestano:
Also note that native mode is only working on Linux with GraalVM 19.3.1. Before it was working on MacOS and with Graal 20 because we didn't needed the substitutions to disable MethodHandles.

I think the MacOS and Graal 20 issues I had will be fixed in upcoming versions of GraalVM

PR #104, opened 2020-05-08
Animal sniffer plugin updates and ignores

pnicolucci:
1) Update the Animal Sniffer Plugin to the latest version as the old version was not working correctly with the `@IgnoreJRERequirement` annotation. 

2) Add the `animal-sniffer-annotations` as a dependency so we can compile with `@IgnoreJRERequirment`

3) Add `@IgnoreJRERequirement` to the methods that use `isSurrogate`

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -1060,9 +1060,9 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.jvnet</groupId>
-                        <artifactId>animal-sniffer</artifactId>
-                        <version>1.2</version>
+                        <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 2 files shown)


tandraschko:
Is optional preffered over scope provided?

wtlucy:
Nice job getting the annotation working.  IMO `<scope>provided</scope>` would make more sense for this, but there must be some historical reason that this file only uses `<optional>true</optional>`?

pnicolucci:
@tandraschko @wtlucy  yes I just kept consistent with what was already in the pom as well as following the instructions for the plugin: https://www.mojohaus.org/animal-sniffer/animal-sniffer-annotations/

pnicolucci:
I'll merge this tomorrow if there are no other concerns. Thanks for the reviews!

PR #105, opened 2020-05-14
add basic file upload

adrianfiedler:
basic example for fileUpload


tandraschko:
works now without commons-fileupload

PR #110, opened 2020-07-04
MYFACES-7347 Rely on BeanManager to determine whether @FlowScoped is passivating.

benkard:
Quarkus does not have passivation and its `BeanManager#isPassivatingScope` implementation therefore always returns `false`.  `FlowScopeBeanHolder` ignores this and assumes that `@FlowScoped` is always passivating (which is presumably true everywhere but Quarkus).  This makes attempts to use `@FlowScoped` beans fail with a `ClassCastException` during `ContextualStorage#getBeanKey`'s attempt to cast the passed bean to `javax.enterprise.inject.spi.PassivationCapable`.

This PR changes `FlowScopeBeanHolder` to not assume that `@FlowScoped` is a passivating scope and to ask the `BeanManager` instead.

This should not affect any CDI implementations other than Quarkus (since `BeanManager.isPassivatingScope(FlowScoped.class)` is expected to return `true` everywhere else), but I have not tested this.

Fixes: [MYFACES-7347](https://issues.apache.org/jira/browse/MYFACES-4347)


tandraschko:
AFAIR i fixed such a casting exception in the past, support passivating and non-passivating in the abstractcontext. Can you check this? ViewScoped should have the same problem and works fine.

NOTE: we already need a JIRA ticket for tracking purpose

benkard:
Regarding `@ViewScoped`:

https://github.com/apache/myfaces/blob/02632b043881dac9de46a48ebd8319c1f3255df9/impl/src/main/java/org/apache/myfaces/cdi/view/ViewScopeContextualStorage.java#L111

`@ViewScoped` has an explicit `instanceof` check, which is another way of fixing the same problem.

tandraschko:
OK
Would be great you can fix it Like viewscoped

benkard:
Sure!  That was actually the first thing I did in my local copy and it worked fine, I just thought it’d be preferable to do a less intrusive thing instead. :)  I’ll update the PR later.

benkard:
Updated.

tandraschko:
Thanks!

PR #112, opened 2020-07-21
Update parent pom to 19

pnicolucci:
When I was building the site for the 2.2.13 release I ran into a number of problems since we're using git now. It was necessary to update the parent pom to 19 locally to get the 2.2.13 tag site to build properly and deployed.

Updating here so we don't have the issue going forward on the next 2.2.x release. However in order to make this change for compilation there were a number of new check styles and the poms/files had to be updated to resolve these issues.

This PR contains all the necessary changes.

1) Update pom version from 16 to 19.
2) Update check style issues found when updating number 1.
3) Updated to use 2.17 version of maven-checkstyle-plugin to be consistent with 2.3/master
4) Add period to javadoc for:
```
@JSFWebConfigParam(defaultValue="false", expectedValues="true, false", since="2.2.13", group="validation")
    private static final String EMPTY_VALUES_AS_NULL_CLEAR_INPUT_PARAM_NAME
            = "org.apache.myfaces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL_CLEAR_INPUT";
```
`ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.3:site (default-site) on project myfaces-impl: Error during page generation: Error parsing '/root/myfaces/myfacesGit22/myfaces/impl/target/generated-site/xdoc/webconfig.xml': line [519] Error parsing the model: end tag name </td> must match start tag name <p> from line 519 (position: TEXT seen ...en INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL is enabled</td>... @519:107) -> [Help 1]`

The problem is that the description for the webconfig param ends at the first period. So in this case the generated files we're messed up b/c the first period came after a `<p>` but before `</p>`

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -82,7 +82,6 @@
                 </executions>
             </plugin>
 
-
             <!-- license checker needs to exclude some kinds of files -->
             <plugin>
                 <groupId>org.apache.rat</groupId>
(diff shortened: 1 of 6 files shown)


tandraschko:
+1

PR #114, opened 2020-08-13
If web.xml defines a faces servlet, take it from there

38leinaD:
To define additional url mappings for example, it should be possible to define the faces servlet via the web.xml.
If it exists in web.xml, it is used for here. If not, a default faces servlet config is created from the extension.


tandraschko:
Thanks for the PR

PR #120, opened 2020-10-12
MYFACES-4360: Faces Config 3.0

volosied:
Let me know if there are any concerns with this PR. Otherwise, I can working some of the issues I found next.

-     We must probably also add the new namespaces to FacesConfigurator.java
The only namespaces I saw were for  http://xmlns.jcp.org/jsf, http://xmlns.jcp.org/jsp, and http://xmlns.jcp.org/jstl. I don't are updated in EE9, but will in EE10. See https://github.com/eclipse-ee4j/faces-api/issues/1526

-     add 3.0 in faces-config-versionType in web-facesconfig_3_0.xsd
Updated to 3.0 

-     we often has stuff like public final static String NAMESPACE = "http://xmlns.jcp.org/jsp/jstl/functions";
Same as first point 

______________
Other items I noticed... 

Questions:  Is _impl/src/main/conf/META-INF/standard-faces-config-base.xml_ update necessary? I believe since this is a new namespace? But 2.3, and 2.2 still reference 2.0 in the other branches. 

Should this be updated to 3.0? Or should we leave it as 2.3? 
```
/Users/siedlecki/open-source/apache/api/src/main/java/jakarta/faces/annotation/FacesConfig.java
Version.JSF_2_3;
```

We should simplify the version checks, and handle for 3.0 as well. 
```
main/java/org/apache/myfaces/config/DefaultFacesConfigurationProvider.java
496:            || "2.2".equals(version) || "2.3".equals(version))
```

I believe the version handing should be improved for 30 as isVersion23OrLater does not seem appropriate. 
```
main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java
236:                            : ("2.3".equals(version) ? FACES_CONFIG_SCHEMA_PATH_23
289:            result = handler.isVersion23OrLater() ? "2.3" : (handler.isVersion22() ? "2.2" :
```

XML_Validation is broken ( I will create a new JIRA issue for it shortly -- https://issues.apache.org/jira/browse/MYFACES-4363)
Validation should should handle jakartaee_9.xsd similar to javaee_5.xsd:
https://github.com/apache/myfaces/blob/8832fa4bf79cbe6cea75e8eaac0a1bc753f0ee25/impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java#L184

### DIFF
--- a/impl/src/main/resources/org/apache/myfaces/resource/web-facesconfig_3_0.xsd
+++ b/impl/src/main/resources/org/apache/myfaces/resource/web-facesconfig_3_0.xsd
@@ -27,7 +27,7 @@
 <xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"
     xmlns:jakartaee="https://jakarta.ee/xml/ns/jakartaee" attributeFormDefault="unqualified"
     elementFormDefault="qualified" targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
-    version="2.3">
+    version="3.0">
 
     <xsd:include schemaLocation="jakartaee_9.xsd"/>
 


tandraschko:
I can see this NS change here: https://github.com/eclipse-ee4j/mojarra/commit/c3b0dc5fcbdc2eca83239d9b2881f19a0e0cbc10

tandraschko:
Maybe you can also check the Mojarra sources of the RC5 branch

bohmber:
Please careful when checking other source  
https://www.apache.org/legal/resolved.html#weak-copyleft-licenses

volosied:
@tandraschko 

Firstly, I apologize for the  delay on this, but I finally spent  some time on this and here's what I found -- 

-  I don't really see anything within MyFaces that  matches the Namespace changes done in Mojarra. 
     - No hits for [NamespaceContext](https://docs.oracle.com/javase/8/docs/api/javax/xml/namespace/NamespaceContext.html?is-external=true) in MyFaces. 

- There doesn't appear to be any Namespace handing within the MyFaces implementation. The only thing that  matters from the faces-config declaration is the [version](https://github.com/apache/myfaces/blob/3.0.x/impl/src/main/java/org/apache/myfaces/config/impl/digester/DigesterFacesConfigUnmarshallerImpl.java#L55)?  I didn't experience any errors or warnings when just using **<faces-config version="3.0">**`.

The code for handling the faces-config isn't very straightforward, but I saw that the parsing is accomplished  by [DigesterFacesConfigUnmarshallerImpl](https://github.com/apache/myfaces/blob/3.0.x/impl/src/main/java/org/apache/myfaces/config/impl/digester/DigesterFacesConfigUnmarshallerImpl.java#L512) which is called from  [here](https://github.com/apache/myfaces/blob/3.0.x/impl/src/main/java/org/apache/myfaces/config/DefaultFacesConfigurationProvider.java#L437). 

Should I first  implement something relating to the NamespaceContext, but I'm not sure if it's necessary?

Last of all, I also don't think this PR needs the impl/src/main/conf/META-INF/standard-faces-config-base.xml change?

tandraschko:
Dont know, to be honest.
I dont think we need such a NamespaceContext if a faces-config with https://jakarta.ee/xml/ns/jakartaee namespace can be parsed by MF.

PR #121, opened 2020-10-20
Myfaces 4364: Fix missing jakarta renames

isaacrivriv:
Fix for https://issues.apache.org/jira/browse/MYFACES-4364


tandraschko:
Thanks!

PR #122, opened 2020-10-30
MYFACES-4363: Correct xsd files for proper validation + Update Version Handling 

volosied:
Fixes MYFACES-4363

Tested locally with the 2.2, 2.3, and 3.0 faces-config version & corresponding namespaces. 

Update: Tests added


tandraschko:
+1 if you added some tests for it

please also add it to master, which is now Jakarta Faces 4.0

volosied:
Added tested :) 
Please let me know if you have issues issues with the version handling updates as well. (I can make it in a separate change instead)

tandraschko:
+1
please also port it to master/4.0
i also refactored the version handling there and defaulted to 4.0 isntead 1.2

tandraschko:
@volosied please merge this to 4.0/master and close the issue

PR #123, opened 2020-11-03
Apply MYFACES-4363 to Master (4.0) 

volosied:
Fixed MYFACES-4363 on Master

Also, added testJSF40Config and impl handling for  jakartaee_10.xsd


tandraschko:
Looks good, thanks!

PR #125, opened 2020-12-11
MYFACES-4372 -- Fix JNDI Name & Add Deprecated Annotations (4.0 Branch)

volosied:
Similar to PR #124

4.0 already has a number of deprecated functions removed, so that's why the number of annotations
 is much smaller. 

Few Notes:
- api/src/main/resources/META-INF/componentClass20.vm already had setValueBinding removed
- StateManager#restoreView was missing deprecated comment
- UIViewRoot#METADATA_FACET_NAME was already properly set

https://issues.apache.org/jira/browse/MYFACES-4372

### DIFF
--- a/api/src/main/java/jakarta/faces/application/ProjectStage.java
+++ b/api/src/main/java/jakarta/faces/application/ProjectStage.java
@@ -30,7 +30,7 @@ public enum ProjectStage
     SystemTest,
     UnitTest;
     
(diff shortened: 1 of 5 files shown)


tandraschko:
+1 both PRs looks ok for me

pnicolucci (approved):

PR #127, opened 2021-01-13
MYFACES-4373: prefer SecureRandom for token generation 3.0.x

wtlucy:
https://issues.apache.org/jira/browse/MYFACES-4373

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/application/viewstate/ServerSideStateCacheImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/application/viewstate/ServerSideStateCacheImpl.java
@@ -133,11 +133,11 @@ class ServerSideStateCacheImpl extends StateCache<Object, Object>
      * Adds a random key to the generated view state session token.
      */
     @JSFWebConfigParam(since="2.1.9, 2.0.15", expectedValues="secureRandom, random", 
-            defaultValue="random", group="state")
+            defaultValue="secureRandom", group="state")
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/apache/myfaces/application/viewstate/ServerSideStateCacheImpl.java
```diff
@@ -133,11 +133,11 @@
      * Adds a random key to the generated view state session token.
      */
     @JSFWebConfigParam(since="2.1.9, 2.0.15", expectedValues="secureRandom, random", 
-            defaultValue="random", group="state")
+            defaultValue="secureRandom", group="state")
     public static final String RANDOM_KEY_IN_VIEW_STATE_SESSION_TOKEN_PARAM
             = "org.apache.myfaces.RANDOM_KEY_IN_VIEW_STATE_SESSION_TOKEN";
     public static final String RANDOM_KEY_IN_VIEW_STATE_SESSION_TOKEN_PARAM_DEFAULT = 
-            RANDOM_KEY_IN_VIEW_STATE_SESSION_TOKEN_RANDOM;
+            RANDOM_KEY_IN_VIEW_STATE_SESSION_TOKEN_SECURE_RANDOM;
```
bohmber, line 140: same here in this file the lines 205-212 if randomMode is unknown

wtlucy: Ah, good catch.. I'll do a new set of PRs to update the "Fallback" to SecureRandom.

PR #130, opened 2021-01-13
MYFACES-4373: prefer SecureRandom for token generation 2.1.x

wtlucy:
https://issues.apache.org/jira/browse/MYFACES-4373


bohmber (approved):
Looks ok for me

PR #133, opened 2021-01-13
MYFACES-4373: make sure SecureRandom is used for invalid configs 3.0.x

wtlucy:
follow-on for https://issues.apache.org/jira/browse/MYFACES-4373


bohmber (approved):

tandraschko:
+1 for both

PR #134, opened 2021-01-13
MYFACES-4373: make sure SecureRandom is used for invalid configs

wtlucy:
follow-on for https://issues.apache.org/jira/browse/MYFACES-4373


bohmber (approved):

bohmber (commented):
Can you apply the change for csrfRandomMode as well?

wtlucy:
> Can you apply the change for csrfRandomMode as well?

Sure, done. I also realized that `StateCacheClientSide` needed to be updated.

bohmber (approved):

PR #135, opened 2021-01-14
addedd github ci actions

bohmber:
Starting with a simple CI run for master. Should be improved in future


wtlucy (approved):
Seems like a good starting place

PR #141, opened 2021-01-14
ci build fixes

bohmber:
Simplify rat config unfortunately we need it twice. Someone added poms with CRLF I need to fix it
You can see the ci build based on github actions is running


wtlucy (approved):
copyright additions and pom line ending changes look fine, and the rat simplification is nice - thanks for this

PR #143, opened 2021-01-14
Only Support Faces Version 3.0  During Validations

volosied:
Supported Faces Versions needs an update somewhere because the current web-facesconfig_3_0.xsd doesn't support version 3.0. 

See https://issues.apache.org/jira/browse/MYFACES-4374

This fix is one option. The other is #144

Code was taken from the https://jakarta.ee/xml/ns/jakartaee/web-facesconfig_3_0.xsd


tandraschko:
Please create a Ticket for Tracking and also apply it for 4.0

tandraschko:
closing in favor of #144

PR #144, opened 2021-01-14
Support Range of Versions During Validations

volosied:
The supported Faces Config versions need an update because the current `web-facesconfig_3_0.xsd `doesn't specifically support version 3.0.

See https://issues.apache.org/jira/browse/MYFACES-4374

This fix is one option.  Otherwise, see #143

### DIFF
--- a/impl/src/main/resources/org/apache/myfaces/resource/web-facesconfig_3_0.xsd
+++ b/impl/src/main/resources/org/apache/myfaces/resource/web-facesconfig_3_0.xsd
@@ -842,6 +842,7 @@
             <xsd:enumeration value="2.1"/>
             <xsd:enumeration value="2.2"/>
             <xsd:enumeration value="2.3"/>
+            <xsd:enumeration value="3.0"/>
         </xsd:restriction>
     </xsd:simpleType>
 


tandraschko:
I would go with this for now
Also please add a unittest and also merge it to master

volosied:
Sounds good. I'll keep the other PR open for later, but go with this one for the 3.0.0 release. Thanks!

wtlucy:
+1 for this PR, to continue supporting old schema versions

volosied:
I'm not sure of a way to add a Unit Test. 

Current stack trace shows --
```
Jan 15, 2021 12:04:51 PM org.apache.myfaces.config.ConfigFilesXmlValidationUtils$ValidationErrorHandler error
SEVERE: cvc-enumeration-valid: Value '3.0' is not facet-valid with respect to enumeration '[1.2, 2.0, 2.1, 2.2, 2.3]'. It must be a value from the enumeration.
org.xml.sax.SAXParseException; lineNumber: 23; columnNumber: 17; cvc-enumeration-valid: Value '3.0' is not facet-valid with respect to enumeration '[1.2, 2.0, 2.1, 2.2, 2.3]'. It must be a value from the enumeration.
	at java.xml/com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:204)
	....
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:824)
	at java.xml/com.sun.org.apache.xerces.internal.jaxp.validation.StreamValidatorHelper.validate(StreamValidatorHelper.java:176)
	at java.xml/com.sun.org.apache.xerces.internal.jaxp.validation.ValidatorImpl.validate(ValidatorImpl.java:115)
	at java.xml/javax.xml.validation.Validator.validate(Validator.java:124)
	at org.apache.myfaces.config.ConfigFilesXmlValidationUtils.validateFacesConfigFile(ConfigFilesXmlValidationUtils.java:248)
	at org.apache.myfaces.config.ConfigFilesXmlValidationUtilsTestCase.testJSF30Config(ConfigFilesXmlValidationUtilsTestCase.java:84)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
.....
```

I was thinking that the `testJSF30Config` should throw an exception, but that doesn't actually occur. 

The error handle just logs the exception instead of re-throwing the exception as in the [fatalError method. ](https://github.com/apache/myfaces/blob/084d3c8d7c21e4c964c1a13b430563bc4bfb7aed/impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java#L226-L228)

https://github.com/apache/myfaces/blob/084d3c8d7c21e4c964c1a13b430563bc4bfb7aed/impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java#L231-L236

Not sure if error handing should changed or not?

PR #146, opened 2021-01-15
Ci fixes 3.0.x

bohmber:


tandraschko:
please also add 2.3-next here, thats the only version currently usable by Quarkus users and its the same codebase as 4.0

PR #148, opened 2021-01-16
prefer secureRandom over random

bohmber:
fixed code warnings


wtlucy:
When I did the earlier PRs I hadn't thought that we needed to update the websocket token as part of myfaces-4373, but we might as well. +1

PR #153, opened 2021-01-19
Disable the parsing of external general entities and external paramet…

bohmber:
…er entities in XML parsing code
MYFACES-4377


tandraschko:
Could this break something?

bohmber:
adding 

`setFeature( "http://apache.org/xml/features/disallow-doctype-decl", true ); 
`

would break.

The rest is from here https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

I checked other apache projects many of them doing the same

tandraschko:
then +1

PR #159, opened 2021-01-24
MYFACES-4376

bohmber:
Update Cryptographic algorithm in StateUtils to a stronger version

(cherry picked from commit 65a0043c5a89b4375496aa295407d8956512e711)


tandraschko:
If we do it for 3.0, then please apply it for 23next too

volosied:
Sorry I didn't see your comment from the master PR. 

From what I read, we definitely should upgrade DES.  

As for HmacSHA256, I don't see any harm in updating. I  found some articles about collisions with SHA1, and SHA256 seems to be generally more preferred.   I'm curious about performance, but users of MyFaces always have the option to change these algorithms if they so choose.   

+1 from me.

volosied (approved):

PR #161, opened 2021-01-25
[Draft] Testing Travis Builds in 3.0.x

volosied:
Trying to exclude jdks in the PRs, so only openjdk8 runs

Otherwise, cron jobs should run daily and build with every jdk listed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,5 +32,6 @@ matrix:
   allow_failures:
     - jdk: openjdk-ea
 
+
 script:
   - mvn clean test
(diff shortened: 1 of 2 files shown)


bohmber:
Can we exclude the linux-ppc64le build as well? 
Hmm as far I can see we should only use the github action for pull requests. We are still waiting for the travis build

volosied:
I can try to remove  linux-ppc64le, too.

Queued, but not started... Let me investigate.

volosied:
Sorry, testing in master since builds aren't running in 3.0.x. Appears to be a seperate issue.

PR #162, opened 2021-01-25
Exclude jdks & linux-ppc64le from Travis CI  PR Builds (master branch)

volosied:
3.0.x Travis Builds are queued and not executing. I'm not sure why, but I'm looking into it. 

For now, I'm testing the exclusions on master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,15 @@ jdk:
   - openjdk-ea
 
 matrix:
+  exclude:
+    - if: type = pull_request
+      jdk: openjdk11
+    - if: type = pull_request
+      jdk: openjdk13
+    - if: type = pull_request
+      jdk: openjdk-ea
+    - if: type = pull_request
+      os: linux-ppc64le
   include:
     - os: linux-ppc64le
       jdk: openjdk8


volosied:
Hmm... Maybe I'm going about it wrong? https://travis-ci.community/t/if-doesnt-appear-to-work-in-exclude/8595/7

volosied:
@bohmber 

From my testing, this `excludes` logic works well. The only issue is that the Apache builds are backed up so these builds may not be executed quickly.   But when executed, it only takes ~5 minutes as compared to ~28 for full JDK/OS set. 

Cron executed today for 3.0.x & master. See https://travis-ci.com/github/apache/myfaces/builds

volosied:
One thing I think need to update is the cron email notifications. Did you receive any today?

bohmber:
Where is the travis email notification configured?

volosied:
I don't see a way to modify notification for cron jobs. 

Only notifications I see are under https://travis-ci.com/account/preferences

 ```
Build emails
The status of your builds straight to your inbox
``` 

However, I think we can have Travis email our myfaces listserv if you'd like (preferably on failures)

bohmber:
It's fine for now. But I think we should use https://ci-builds.apache.org/job/MyFaces/ instead of Travis

bohmber:
Tobago is using a pipline

https://ci-builds.apache.org/job/MyFaces/job/Tobago%20pipeline/

https://github.com/apache/myfaces-tobago/blob/master/Jenkinsfile

volosied:
I was wondering how some people made those pipelines. Let me try to add something for us as well.

PR #191, opened 2021-03-30
MYFACES-4390: HtmlInputFile multiple support

melloware:
I feel like an alien in this code base!

I didn't see any existing tests to update so i wasn't even sure where to begin to add a test for this?

### DIFF
--- a/api/src/main/java/jakarta/faces/component/html/_HtmlInputFile.java
+++ b/api/src/main/java/jakarta/faces/component/html/_HtmlInputFile.java
@@ -83,6 +83,16 @@ abstract class _HtmlInputFile extends UIInput implements _AccesskeyProperty,
     @JSFProperty
     public abstract String getAccept();
 
+    /**
+     * Flag indicating that this element must allow multiple file selection. A value
+     * of false causes no attribute to be rendered, while a value of true causes the attribute to be rendered as
+     * multiple="multiple".
+     *
+     * @return the multiple value
+     */
+    @JSFProperty(defaultValue="false")
+    public abstract boolean isMultiple();
+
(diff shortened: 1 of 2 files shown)


melloware:
Also there were some warnings from IntelliJ about some evaluations ALWAYS being true etc but I did not fix any of them only touched the bare minimum to add this feature.

melloware:
Also unsure as its mentioned it should render `multiple="multiple"` but right now I think it would render `multiple="true"`.  Wasn't sure how to do that in MyFaces with JSFProperty.

tandraschko:
why not
```writer.writeAttribute(HTML.MULTIPLE_ATTR, "multiple", null);``` then?

**Comment by melloware** [comment]
OK I updated it. to write out "multiple" but what threw me off is the lines above it were this...
```java
if (inputFile.isDisabled())
{
     writer.writeAttribute(HTML.DISABLED_ATTR, Boolean.TRUE, null);
}
```

But in PrimeFaces code we do this...
```java
 writer.writeAttribute("disabled", "disabled", null);
```

GedMarc:
yeah disabled=true doesn't work across all browsers, has to be  just "disabled" or "disabled=disabled"

GedMarc:
ref here https://stackoverflow.com/questions/6961526/what-is-the-correct-value-for-the-disabled-attribute

Don't use attributes=true for these at all, ever

melloware:
Thanks @GedMarc. So @tandraschko so then the Disabled attribute here is incorrect then.

tandraschko:
@melloware would you please add a jira issue and provide a PR? would be great!

PR #198, opened 2021-04-06
MYFACES-4393: Render type='text/javscript` only on HTML4

melloware:
- removed LANGAUGE constants as they were not used anywhere in the code base and are also legacy attribute
- Fixed all places that were delcaring type='text/javascript` to only render if its not HTML5 doctype


tandraschko:
looks good, thanks!

PR #199, opened 2021-04-06
CI Java v2 and Adopt OpenJDK

melloware:
Fixed line endings of the one file. Updated to Actions V2 and Adopt OpenJDK


tandraschko:
i think we may need to adjust branches as well?

melloware:
I will submit PR's for the other branches as well

bohmber (approved):

PR #211, opened 2021-06-23
MYFACES-4404 -- Log viewstate string

volosied:
https://issues.apache.org/jira/browse/MYFACES-4404

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java
+++ b/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java
@@ -20,7 +20,9 @@ package org.apache.myfaces.renderkit.html;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+
 
 import javax.faces.context.ExternalContext;
 import javax.faces.context.FacesContext;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java
```diff
@@ -121,6 +121,9 @@ private void writeViewStateField(FacesContext facesContext, ResponseWriter respo
     {
         String serializedState = getStateCache(facesContext).getStateTokenProcessor(facesContext)
                 .encode(facesContext, savedState);
+
+        log.info("Writing serialized viewstate string: " + serializedState);
```
melloware, line 125: This will always concat the string shouldn't you wrap this in something like...

```
if (log.isInfoEnabled()) { log.info() )} 
```

wtlucy (approved):
LGTM

impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java
volosied: Good point!

melloware (approved):

tandraschko:
TBH i dont like that
viewstate string can be very big with client side state saving and therefore you will spam the logfiles

volosied:
We already log it here: https://github.com/apache/myfaces/blob/584d6197a25ab4fdc24aaba5fb00b29ba14d2d0c/shared/src/main/java/org/apache/myfaces/shared/util/StateUtils.java#L394

Maybe I could log the hash instead?

volosied:
Logging would only be enabled when debugging, so the line could still be useful.

PR #221, opened 2021-07-08
Apply MYFACES-4404 to Master 

volosied:
Also changed CRLF with LF for some files.


impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlResponseStateManager.java
```diff
@@ -113,6 +113,11 @@ private void writeViewStateField(FacesContext facesContext, ResponseWriter respo
         String serializedState = getStateCache(facesContext).getStateTokenProcessor(facesContext)
                 .encode(facesContext, savedState);
 
+        if (log.isLoggable(Level.FINE))
```
wtlucy, line 116: you need to import this `java.util.logging.Level`

wtlucy (approved):

PR #228, opened 2022-02-07
MYFACES-4428 - Separate "begin tag" code from "end tag" code

exabrial:
MYFACES-4428 - Separate "begin tag" code from "end tag" code

JSF provides a encodeBegin() and an encodeEnd() method on a Renderer. MyFaces currently performs "begin tag" and "end tag" encoding in the encodeEnd(), which breaks any vendor-netural path for extensibility. This patch moves the "begin tag" encoding logic to the encodeBegin() function so the renderkit can be extended without needed to extend myfaces classes explicitly, but instead be called through the standard JSF api to start a tag.

### DIFF
--- a/impl/src/test/java/org/apache/myfaces/renderkit/html/HtmlSecretRendererTest.java
+++ b/impl/src/test/java/org/apache/myfaces/renderkit/html/HtmlSecretRendererTest.java
@@ -77,6 +77,7 @@ public class HtmlSecretRendererTest extends AbstractJsfTestCase
 
     public void testInputTextDefault() throws Exception
     {
+        inputText.encodeBegin(facesContext);
         inputText.encodeEnd(facesContext);
         facesContext.renderResponse();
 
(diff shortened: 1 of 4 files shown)


melloware:
This seems like a pretty big change to me.  I totally get what you are saying but let's see what @tandraschko says 😁

tandraschko:
build fails :D

exabrial:
I can work on fixing the test, it looks like it's expecting the encodeEnd to render the begin tag.

exabrial:
apologies in advance for asking a stupid question: how do I re-run the build on Github....?

melloware:
I just kicked it off again.  Its because you are a first time contributor that GitHub blocks the actions from running automatically.

melloware:
Build succeeded!

exabrial:
Yay! we've tried this out (TomEE 8.0.9) on a couple of internal, non-trivial JSF apps without issue.

tandraschko:
looks good to me
can you also do this for 2.3-next, 3.0 and 4.0?
we need to keep those synced

volosied:
Hi, I also ran this change on on my team's suite of tests. Everything looked good.

exabrial:
I pushed PRs for those additional branches. Note the master branch does not pass RAT check locally

PR #233, opened 2022-03-10
JSF provides a encodeBegin() and an encodeEnd() method on a Renderer.…

glynn29:
… MyFaces currently performs "begin tag" and "end tag" encoding in the encodeEnd(), which breaks any vendor-neutral path for extensibility. This patch moves the "begin tag" encoding logic to the encodeBegin() function so the renderkit can be extended without needed to extend myfaces classes explicitly, but instead be called through the standard JSF api to start a tag.

### DIFF
--- a/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlCheckboxRendererBase.java
+++ b/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlCheckboxRendererBase.java
@@ -54,7 +54,8 @@ public class HtmlCheckboxRendererBase extends HtmlRenderer
 
     private static final String EXTERNAL_TRUE_VALUE = "true";
 
-    public void encodeEnd(FacesContext facesContext, UIComponent uiComponent)
+    @Override
+    public void encodeBegin(FacesContext facesContext, UIComponent uiComponent)
             throws IOException
     {
         org.apache.myfaces.shared.renderkit.RendererUtils.checkParamValidity(facesContext, uiComponent, null);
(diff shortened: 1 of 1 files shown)


melloware:
Wasn't this already done in this PR?

https://github.com/apache/myfaces/pull/228

glynn29:
It is very similar, but for a different base renderer class. This one is for HtmlCheckboxRendererBase, the other was for HtmlTextRendererBase.

melloware:
cc @exabrial

exabrial:
@glynn29 actually works at the same company as I do :) We ran into the same issue with another renderer. We're going to try and fix these in the same manner as #228 as we find them.

melloware (approved):

glynn29:
Hello, I have posted a backport PR for each version and was checking to see if there were any changes or improvements i can make to them. Thank you!

glynn29:
Thank you for the feedback, I'm hoping my following explanation helps. Perhaps I should add the following as comments? 

This renderer actually handles two classes: UISelectBoolean and UISelectMany. We are merely trying to fix UISelectBoolean in this patch, and not UISelectMany at this time to try to keep merge requests small and focused. 

The single checkbox INPUT_ELEM for UISelectBoolean is opened in `encodeBegin()`, and closed in the `encodeEnd()`. 

The INPUT_ELEM for UISelectMany is left alone in `encodeEnd()`. In `renderCheckboxList()` for UISelectMany, it calls the single checkbox renderer `renderCheckbox()`. The additional branch to check for an instanceof UISelectMany leaves the the current behavior for writing the end element unchanged for this class. 

We can make changes to handle both classes in this patch, but we'd prefer to handle in a separate PR. 

We have tested this manually and it is passing the unit tests written in `HtmlCheckboxRendererTest`. 

Hopefully this clears things up. Thank you!

melloware:
@glynn29 thanks for your excellent contribution!  I am going to merge these and then I am going to try and write a unit test that checks the output of this component. I noticed there are no real unit tests for HTML components in this suite so its a gap.

exabrial:
@melloware we're willing to help out and write a few tests if you can give us some direction on what you'd like to see :) tag me and I'll see what we can whip up

melloware:
@exabrial awesome.  It looks like there are some tests for the components there.  From Thomas "if we have like 2-3 unittests for each component (UISelectMany  and UISelectBoolean), which also checks the rendered markup"

So I think his idea was capturing and verifying the output of the Response Writer!

PR #238, opened 2022-04-01
Fix NPE - MYFACES-4431

volosied:
https://issues.apache.org/jira/browse/MYFACES-4431


melloware:
@tandraschko just checking does this need to be ported to 3.X or 4.X branches?

tandraschko:
yep, please!

melloware:
OK will do!

PR #255, opened 2022-04-27
MYFACES-4433 - Fix ViewScopeContextualStorage memory leak

volosied:
https://issues.apache.org/jira/browse/MYFACES-4433

I believe we still want to call ViewScopeContextImpl.destroyAllActive(st); in order to invoke PreDestroy on the ViewScoped beans. 

Once that's done, we can remove the contextualStorage from the storageMap


melloware:
This looks good to me but I prefer @tandraschko give it his blessing :)

melloware (approved):

impl/src/main/java/org/apache/myfaces/cdi/view/ViewScopeBeanHolder.java
```diff
@@ -236,4 +236,8 @@ public String generateUniqueViewScopeId()
         return key;
     }
 
+    public void removeStorage(String viewScopeId){
```
volosied, line 239: Maybe rename this to `removeStorageContextFromMap` instead?  

Let me know if you have a preference or any other suggestions.

melloware: I think `removeStorage` is good.  Clean and concise

tandraschko:
Does it only exist in 2.2?

tandraschko:
But in general it looks good

melloware:
I assume this needs to be forward ported to the higher versions?

volosied:
It looks like this affects 2.3 and 2.3-next, too. I'll get PRs for them up soon.

4.0 was refactored, so I'm not sure if it is a problem there anymore?

tandraschko:
if 2.3 is affected, then also 3.0
and i think if 2.3-next is affected, also 4.0 should be (its a very similar codebase)

melloware:
@volosied saw your votes for 2.3, next and 3.0.  Any chance you could do a 2.2 release as well?  Maybe it will be the last one but it fixes a few significant things.

volosied:
@melloware  Sure, I'll do that later today or tomorrow at the latest :)

PR #259, opened 2022-04-29
MYFACES-4434 - HtmlTextareaRendererBase separate encode end and begin

glynn29:
… MyFaces currently performs "begin tag" and "end tag" encoding in the encodeEnd(), which breaks any vendor-neutral path for extensibility. This patch moves the "begin tag" encoding logic to the encodeBegin() function so the renderkit can be extended without needing to extend myfaces classes explicitly, but instead be called through the standard JSF api to start a tag.


melloware (approved):

melloware:
Looks good!   As usual if you don't mind submitting PR's for other branches :)  Thanks!

glynn29:
Thank you, have created 3/4 of the PRs. Will submit the last one on moday with a fellow co-worker.

melloware:
Thanks I will review them all but this looks pretty straightforward!

melloware:
I took care of adding the 3.0.x branch change.

PR #263, opened 2022-05-03
MYFACES-4435 - Update Language 2.3-next

volosied:
[MYFACES-4435](https://issues.apache.org/jira/browse/MYFACES-4435)

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientConfig.java
+++ b/impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientConfig.java
@@ -198,7 +198,7 @@ public class ClientConfig implements Serializable
     /**
      * Users can overload this method to define in which scenarios a request should result
      * in an 'intercepted' page with proper windowId detection. This can e.g. contain
-     * blacklisting some userAgents.
+     * blocklisting some userAgents.
      * By default the following User-Agents will be served directly:
      * <ul>
      *     <li>.*bot.*</li>
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/_util/_Lang.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/_util/_Lang.js
@@ -250,7 +250,7 @@ _MF_SINGLTN(_PFX_UTIL + "_Lang", Object, /** @lends myfaces._impl._util._Lang.pr
      * @param {Object} src the source map
      * @param {boolean} overwrite if set to true the destination is overwritten if the keys exist in both maps
      **/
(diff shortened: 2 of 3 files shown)


impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientConfig.java
```diff
@@ -198,7 +198,7 @@ public String getUserAgent(FacesContext facesContext)
     /**
      * Users can overload this method to define in which scenarios a request should result
      * in an 'intercepted' page with proper windowId detection. This can e.g. contain
-     * blacklisting some userAgents.
+     * blocklisting some userAgents.
```
melloware, line 201: isn't blacklisting the correct word here?

melloware (changes requested):

impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientConfig.java
volosied: Both mean the same thing, but the connotations are different.  It relates to blacklisting/whitelisting, which can be viewed via a racial lens.

melloware: Got it!

melloware (approved):

impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientConfig.java
mkienenb: "blocklisting" is not a word.   If you want to replace "blacklist", then pick a real word or phrase to do so.   Not everyone speaks English as a first language, and if you start picking words that don't appear in dictionaries, you will be doing far more excluding than including.

melloware: Yes I would suggest rewording it rather than using a non-existent word.

melloware:
Just to close the loop on this I found this article:  https://www.vice.com/en/article/v7dd3d/we-need-to-stop-saying-blacklist-and-whitelist

> The developers of Chromium, the open source codebase that underpins Google’s Chrome browser, as well as the competing Microsoft’s Edge, [announced in 2019](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/inclusive_code.md) that “terms such as ‘blacklist’ and ‘whitelist’ reinforce the notion that black==bad and white==good.” That’s why since then its developers are using “blocklist” and “allowlist.”

Basically good enough for Google the word `blocklist` will become the new word 😄

PR #269, opened 2022-06-09
MYFACES-4417: Cookie SameSite=Strict

melloware:
Found two cookies that should be set to Strict.

1. The Flash cookie
2. The cookie that disables/enables JavaScript.


tandraschko:
+1

PR #270, opened 2022-06-10
MYFACE-4387: JSF to Faces in Javadoc and Constants

melloware:
@volosied @tandraschko you definitely don't have to accept this but I renamed most JSF to Faces in Javadoc and Consttants. etc.

The one thing I did not touch is the Build Tools annotations like `@JSFComponent(name="ui:debug")` etc since those build tools are still used across all versions.


tandraschko:
We even cant merge it
It changes API enums and removes constants
This must all be done in spec first

melloware:
no problem i will close it

volosied:
Thanks for looking into this so quick. 

Should I mark MYFACE-4387 as fixed in 4.0.0-RC1 then?

melloware:
This should be marked for 4.0.0-RC2 since it just recently got merged!

PR #271, opened 2022-06-12
MYFACE-4387: JSF to Faces in Javadoc

melloware:
@tandraschko this change is just Javadoc only i didn't touch any Java code or constants at all


api/src/main/java/jakarta/faces/component/UIViewParameter.java
```diff
@@ -45,7 +45,7 @@
  * TODO: documentation on jsp and pld are not the same. It appear two
  * params: maxlength and for, but no property getter and setter founded here. 
  * If maxlength is used, we can put something like this: 
- * JSFJspProperty(name = "maxlength", returnType = "java.lang.String")
```
lofwyr14, line 48: This is an annotation. AFAIK the class names will not be renamed.

impl/src/main/java/org/apache/myfaces/view/facelets/tag/faces/html/DefaultTagDecorator.java
```diff
@@ -176,7 +176,7 @@ public Tag decorate(Tag tag)
             return null;
         }
         
-        // One or many attributes has the JSF_NAMESPACE attribute set. Check empty or
+        // One or many attributes has the Faces_NAMESPACE attribute set. Check empty or
```
lofwyr14, line 179: Here JsfLibrary.NAMESPACE is correct. JSF_NAMESPACE was replaced.

lofwyr14 (changes requested):
Found 2 code references that can't be changed in this was.

Discussion: If there is a text like e.g.: "See JSF 1.2" 
I would not rename it to "See Faces 1.2", because the versions of JSF 1 and 2 will not be renamed. This is history and can't be "rewritten".

melloware:
Thank you both changes made!

tandraschko:
Can we merge?

melloware:
Its good from mine end.

PR #274, opened 2022-06-24
java 8 improvements:

arturobernalg:
* Replace Anonymous type  with lambda
* Simplifiable 'Map' operations
* Replace Loop with 'Collection.removeIf()'


tandraschko:
Thanks for the PR but that are cosmetics and everyone has other preferences 

I will likely partly apply some of those changes
Please discuss first next time before investing time

PR #277, opened 2022-06-24
Simplify conditional expressions.

arturobernalg:


tandraschko:
Thanks for the PR but that are cosmetics and everyone has other preferences

PR #280, opened 2022-06-25
Reuse expression and avoid duplicate code.Easier to maintain.

arturobernalg:


impl/src/main/java/org/apache/myfaces/view/facelets/compiler/SAXCompiler.java
```diff
@@ -631,14 +631,15 @@ public void endDTD() throws SAXException
         @Override
         public void endElement(String uri, String localName, String qName) throws SAXException
         {
+            final boolean nameSpace = CompositeLibrary.NAMESPACE.equals(uri)
```
melloware, line 634: Maybe call the variable `hasNameSpace` or `isInNameSpace` something more explicit that its a boolean?

melloware (changes requested):

impl/src/main/java/org/apache/myfaces/view/facelets/compiler/SAXCompiler.java
arturobernalg: changed.

tandraschko:
i applied your changes but giving it better var names
thanks!

PR #281, opened 2022-06-25
Use try-with-resources Statement and avoid closing it manually.

arturobernalg:


impl/src/main/java/org/apache/myfaces/config/impl/FacesConfigUnmarshallerImpl.java
```diff
@@ -179,18 +180,8 @@ public FacesConfigImpl getFacesConfig(InputStream in, String systemId) throws IO
...
-            {
-                in.close();
-            }
-            catch (IOException e)
-            {
-                // ignore silently
-            }
-        }
-          
+        // ignore silently
```
melloware, line 183: I don't know if you need these trailing comments?

impl/src/main/java/org/apache/myfaces/resource/TempDirFileCacheContractResourceLoader.java
```diff
@@ -229,20 +227,7 @@ protected void createTemporalFileVersion(FacesContext facesContext, ResourceMeta
...
-                {
-                    inputStream.close();
-                }
-                catch (IOException e)
-                {
-                    // Ignore
-                }
-            }
-        }
+        // Ignore
```
melloware, line 230: Same?

melloware (changes requested):

tandraschko:
also it would throw a IOException when close fails? currently we ignore it

impl/src/main/java/org/apache/myfaces/config/impl/FacesConfigUnmarshallerImpl.java
arturobernalg: done

impl/src/main/java/org/apache/myfaces/resource/TempDirFileCacheContractResourceLoader.java
arturobernalg: done

arturobernalg:
> also it would throw a IOException when close fails? currently we ignore it

Well, changed. but in the class TempDirFileCacheContractResourceLoader its not possible without break the compatibility.

tandraschko:
We dont use final
Also we should not throw a facesexception

tandraschko:
I think we should not touch all cases where close ist silently ignore

arturobernalg:
> I think we should not touch all cases where close ist silently ignore

Now I'm lost.

tandraschko:
there is no nice way to silnently ignore ioexception only on #close in try-resource
therefore i just pushed: https://github.com/apache/myfaces/commit/b79b08ece301f2959332df0b950c4bcc712f360e

PR #282, opened 2022-06-29
MYFACES-4432 - Resolve request object (Ports MYFACES-4394)

volosied:
This was unfortunately missed.  Will be included in the next 3.0.3 release.


melloware (approved):

melloware:
Something to discuss but once 4.0 goes final we should stop making changes to 3.x as no one really uses it.  Our two active branches should just be 2.3 and 4.0 after that.   Just my two cents...and its seems to be the way Mojarra is going as well!

volosied:
@melloware i see you point, but it's one of our product's supported features.  Any bug fixes to 2.3 would be nice to port to 3.0 (vice versa) as well for us.

tandraschko:
As well as 2.3next for quarkus users

PR #288, opened 2022-08-24
MYFACES-4442: Fix UnsupportedOperationException during Faces init

pnicolucci:
This fixes MYFACES-4442: https://issues.apache.org/jira/browse/MYFACES-4442

I had to rummage around in the commit history of the current branches and find where these classes used to reside:

1. [ServletMapping](https://github.com/apache/myfaces/commit/ae3da459a80115650e81ca89d870183f0c9c434b#diff-38135243715d25b010f65297f26b41e66cc7205a2d6b9672407fc5d52c301bbb)
2. [WebXml](https://github.com/apache/myfaces/commit/115e2e3a3817e9fdf091a8672dc3bfd58a02ecdb#diff-65b3eb07dd1234eefee5b611904e4af48a2d4a53043ca72c7377e00c37d774ae) I updated this class to correct javax->jakarta, ClassUtils current package, updated import for DelegatedFacesServlet, and updated package for WebXmlParser.
3. [WebXmlParser](https://github.com/apache/myfaces/blob/3.0.x/impl/src/main/java/org/apache/myfaces/shared_impl/webapp/webxml/WebXmlParser.java) Already existed. [History of WebXmlParser](https://github.com/apache/myfaces/commits/main/impl/src/main/java/org/apache/myfaces/util/WebXmlParser.java) I added the following back to it and modified it as necessary to fit into the latest code base: [This is what it looked like before it was moved in the same commit](https://github.com/apache/myfaces/commit/ca63ed50afd5caec20f816864cbce196013e1cbd). The readWebApp method was updated to read the different mappings , which was removed [here](https://github.com/apache/myfaces/commit/ae3da459a80115650e81ca89d870183f0c9c434b).
4. [WebXmlParserErrorHandler](https://github.com/apache/myfaces/commit/ca63ed50afd5caec20f816864cbce196013e1cbd)
5. [WebXmlParserUtils](https://github.com/apache/myfaces/commit/ae3da459a80115650e81ca89d870183f0c9c434b)

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/spi/WebConfigProvider.java
+++ b/impl/src/main/java/org/apache/myfaces/spi/WebConfigProvider.java
@@ -18,7 +18,9 @@
  */
 package org.apache.myfaces.spi;
 
+import java.util.List;
 import jakarta.faces.context.ExternalContext;
+import org.apache.myfaces.util.ServletMapping;
 
 /**
  * SPI to provide a custom WebConfigProvider implementation.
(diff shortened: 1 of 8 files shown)


tandraschko:
I think the fallback should be directly in FacesServletMappingUtils.getFacesServletRegistration

AFAICS we dont scan web-fragments? Thats a must have

pnicolucci:
@tandraschko  I have another idea and I think we can avoid this web.xml parsing altogether. I'll work to get a new PR to share my findings and we can discuss.

PR #292, opened 2022-08-29
MYFACES-4442: avoid UnsupportedOperationException

pnicolucci:
Fix for MYFACES-4442: https://issues.apache.org/jira/browse/MYFACES-4442

**Current behavior:**
1) MyFacesContainerInitializer.onStartup
	a) Checks clazzes from @HandlesTypes
	b) Checks if a faces-config.xml is present.
	c) If a or b
		- Check if there are current FacesServlet mappings
		- If there is a  FacesServlet defined set: FACES_SERVLET_FOUND(true)
		- If there is not a FacesServlet defined, then dynamically add and set: FACES_SERVLET_ADDED_ATTRIBUTE(true)

2) StartupServletContextListener.contextInitialized
	a) Call FacesInitializer.initFaces(ServletContext)

3) FacesInitializerImpl.initFaces
	a) If !INITIALIZE_ALWAYS_STANDALONE
	b) Use FacesServletMappingUtils to getFacesServletRegistration
	c) if getFacesServletRegistration is null / 0 then perform other checks
	d) Check MyFacesContainerInitializer.FACES_SERVLET_FOUND
	c) If MyFacesContainerInitializer.FACES_SERVLET_FOUND is null/false then check FACES_SERVLET_ADDED_ATTRIBUTE
	d) If FACES_SERVLET_ADDED_ATTRIBUTE is null / false then put out message: "No mappings of FacesServlet found. Abort  
             initializing MyFaces."

**Proposed Solution:**
1) MyFacesContainerInitializer.onStartup
	a) set isFacesServletPresent if there is a current FacesServlet defined.
	b) Checks clazzes from @HandlesTypes
	c) Checks if a faces-config.xml is present.
	d) If b or c
		- If there is not a FacesServlet defined (!isFacesServletPresent), then dynamically add and set: 
		FACES_SERVLET_ADDED_ATTRIBUTE(true)

2) StartupServletContextListener.contextInitialized
	a) Call FacesInitializer.initFaces(ServletContext)

3) FacesInitializerImpl.initFaces
	a) If !INITIALIZE_ALWAYS_STANDALONE
	d) Check MyFacesContainerInitializer.FACES_SERVLET_FOUND
	c) If MyFacesContainerInitializer.FACES_SERVLET_FOUND is null/false then check FACES_SERVLET_ADDED_ATTRIBUTE
	d) If FACES_SERVLET_ADDED_ATTRIBUTE is null / false then put out message: "No mappings of FacesServlet found. Abort 
         initializing MyFaces."

Notes:
I don't see any reason why we would need to parse the web.xml or look do another lookup for a FacesServlet in the FacesInitializerImpl when the FacesServlet was already looked up in the MyFacesContainerInitializer. Also if a FacesServlet is not defined and @HandlesTypes classes are passed into onStartup or there is a faces-config.xml available in the application
then the FacesServlet is added dynamically. So for any application that is "Faces enabled" we would either already
have a FacesServlet defined in the web app or we'd try to dynamically add a FacesServlet in onStartup.

To me this pre-Faces 4.0 web.xml parsing and additional lookup seem like old legacy code that perhaps has been around since before ServletContainerInitializers were available (before Servlet 3.0).

This will avoid the UnsupportedOperationException from initFaces for calling getServletRegistrations from a
programmatically added listener: StartupServletContextListener.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/webapp/FacesInitializerImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/webapp/FacesInitializerImpl.java
@@ -85,19 +85,19 @@ public class FacesInitializerImpl implements FacesInitializer
     private static final Logger log = Logger.getLogger(FacesInitializerImpl.class.getName());
 
     public static final String CDI_BEAN_MANAGER_INSTANCE = "oam.cdi.BEAN_MANAGER_INSTANCE";
-    
+
     private static final String CDI_SERVLET_CONTEXT_BEAN_MANAGER_ATTRIBUTE = 
         "jakarta.enterprise.inject.spi.BeanManager";
 
     public static final String INJECTED_BEAN_STORAGE_KEY = "org.apache.myfaces.spi.BEAN_ENTRY_STORAGE";
 
     public static final String INITIALIZED = "org.apache.myfaces.INITIALIZED";
-    
+
     private static final byte FACES_INIT_PHASE_PREINIT = 0;
     private static final byte FACES_INIT_PHASE_POSTINIT = 1;
     private static final byte FACES_INIT_PHASE_PREDESTROY = 2;
     private static final byte FACES_INIT_PHASE_POSTDESTROY = 3;
-    
+
     /**
      * Performs all necessary initialization tasks like configuring this Faces
      * application.
@@ -115,14 +115,14 @@ public class FacesInitializerImpl implements FacesInitializer
             }
             return;
         }
-        
+
         try
         {
             if (log.isLoggable(Level.FINEST))
             {
                 log.finest("Initializing MyFaces");
             }
-            
+
             long start = System.currentTimeMillis();
 
             // Some parts of the following configuration tasks have been implemented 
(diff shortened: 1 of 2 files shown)


tandraschko:
i like this fix but this doesnt solve https://issues.apache.org/jira/browse/TOMEE-2846, which is actually related to the same root problem

pnicolucci:
@tandraschko  I opened https://issues.apache.org/jira/browse/MYFACES-4447 to address automatic extensionless mapping.

This fix at least gets us past the initial initFaces exception and allows me to be unblocked and start investigating other Faces 4.0 issues that we have encountered.

pnicolucci:
So I'll plan to merge this by the end of the day today if no objections to the actual code change here except for we still need to fix it for automatic extensionless mapping which has been broken since 2.3 in this scenario it looks.

melloware (approved):

pnicolucci:
Opened: to fix extensionless mapping: https://issues.apache.org/jira/browse/MYFACES-4447

volosied (approved):
+1 

I tested a few scenarios with this code and everything looked good.

PR #299, opened 2022-08-31
MYFACES-4449

pnicolucci:
Fixes MYFACES-4449: https://issues.apache.org/jira/browse/MYFACES-4449

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/util/lang/ClassUtils.java
+++ b/impl/src/main/java/org/apache/myfaces/util/lang/ClassUtils.java
@@ -19,11 +19,15 @@
 package org.apache.myfaces.util.lang;
 
 import jakarta.faces.FacesException;
+import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.net.URL;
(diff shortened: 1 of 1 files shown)


melloware (approved):

tandraschko:
@pnicolucci 
could you please add something to each method like 
`//@Overide because ClassLoader blalalala`

impl/src/main/java/org/apache/myfaces/util/lang/ClassUtils.java
```diff
@@ -185,6 +189,99 @@ public static <T> T wrapBackwardCompatible(Class<T> interfaceClass, Class<? exte
...
+     * the ClassUtils class loader (i.e. the loader of the myfaces.jar lib) if necessary.
+     *
+     * @param type
+     *            fully qualified name of a non-primitive non-array class
+     * @return the corresponding Class
+     * @throws NullPointerException
+     *             if type is null
+     * @throws ClassNotFoundException
+     */
+    public static <T> Class<T> classForName(String type) throws ClassNotFoundException
```
volosied, line 203: Would you want to add similar comments as in 4.0?

PR #305, opened 2022-09-02
MYFACES-4454

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4454


volosied (approved):
looks good, successful built locally

tandraschko:
yeah
a shame that we have to revert but...

PR #312, opened 2022-09-15
Myfaces 4454 jsf23 next

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4454

The other change for this in 4.0 doesn't look to exist in 2.3-next: https://github.com/apache/myfaces/blob/2.3-next/api/src/main/java/javax/faces/render/Renderer.java#L144 (still uses UIComponent correctly vs T)

Faces 4.0 PR: https://github.com/apache/myfaces/pull/305


tandraschko:
Paul, i dont think 4.0 API changes should be in 2.3-next

pnicolucci:
> Paul, i dont think 4.0 API changes should be in 2.3-next. Its still 2.3 API

I opened the PR against the wrong branch, not sure how I missed that. Sorry!

PR #318, opened 2022-09-23
MYFACES-4456: Add type to HtmlInputText

melloware:
Adds type to HtmlInputText and adds unit tests proving the default is still `type="text"` but can now be set properly on the tag like `type="tel"` etc.

### DIFF
--- a/api/src/main/java/jakarta/faces/component/html/_HtmlInputText.java
+++ b/api/src/main/java/jakarta/faces/component/html/_HtmlInputText.java
@@ -38,7 +38,7 @@ abstract class _HtmlInputText extends UIInput implements _AccesskeyProperty,
     _AltProperty, _UniversalProperties, _DisabledReadonlyProperties,
     _FocusBlurProperties, _ChangeProperty, _SelectProperty, 
     _EventProperties, _StyleProperties, _TabindexProperty, _LabelProperty, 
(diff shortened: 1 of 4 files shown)


pnicolucci:
Looking at the vld doc: https://jakarta.ee/specifications/faces/4.0/vdldoc/ I see 
it states the following:

> Controls the data type and associated control of the element. Default value is text. 
> If a value is specified for which already a more specific component exist, such as 
> hidden, password, checkbox, radio, file, submit, image, reset and button, and the 
> ProjectStage is set to Development, then the renderer must add a warning message 
> stating that the more specific component should be preferred.

melloware:
I saw that but felt like it was overkill to warn the user?  I see Mojarra is doing that but not sure that is really necessary?

melloware:
I can add that though.

tandraschko:
-0 for adding such a warning
there would be much more useful warnings like https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-3127?filter=allopenissues

pnicolucci:
I don't feel strongly either way, just wanted to note what the doc stated. If the community consensus is to not add it I'm fine with that.

melloware:
OK warning added exactly like Mojarra does it.

PR #321, opened 2022-09-24
MYFACES-4465: avoid ClassCastException

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4465


melloware (approved):

tandraschko:
how can this happen? never had this problem in 2.3-next, which is used by some users

pnicolucci:
I've encountered this one time so far with Java11 and many times on Java 17. I'll work to get some issues opened so we can try and take a look at what's not working as expected when using the LambdaMetaFactory. However, merging this at least lets the fall back scenario function correctly.

tandraschko:
Then lets merge

PR #325, opened 2022-09-29
Feature/myfaces 4456 - New faces.js codebase (target 4.0.0 RC3)

werpu:
Myfaces faces.js reimplementation via typescript for MyFaces 4.0, please review


melloware:
Nice work on this!  Lots to review here but looking good from what I see so far.

werpu:
Yes I started with it 2019, but given customer projects, covid and months
of rehab due to a spine injury it took longer than expected.
I reviewed my readmes today, and there was the sentence, that I expect a
final version late 2020, oh boy I was wrong.

Just a short explanation to the build process, the code is now entirely
hosted on myfaces the build dependency for the client less so. But I did
not want to omit the path from the other github project into myfaces, since
it contains a self running version and a 2.3 jsf version. So I pull
basically per user command, not build command, the code via an npm package
in and then copy the sources into myfaces. The rest is a standard npm
webpack build with all the needed dependencies triggered via the client
plugin.

The final file then lands directly in the target folder, also the jsdoc
profile still is intact, it is now just not generated by our jsdoc
generators anymore but the webpack plugin.
(mojarra seems to use the same engine, as it seems, given the html layout
of the jsdoc specs)

I guess apache has already some npm caches in place, because we are not the
first project to do so.

Also another thing, mapping files now work, but wont work out of the box
given how they are loaded, following trick does it:


<servlet-mapping>

...

    
    
    <url-pattern>*.mf_map</url-pattern>

</servlet-mapping>

This adds a special entry which then should load the mapping files, that
way you can debug directly into the ts files if you want to.
Thats the reasony why i was able to kill the 5 different builds for
debugging.


Werner


Am Do., 29. Sept. 2022 um 14:13 Uhr schrieb Melloware <
***@***.***>:

melloware (approved):

volosied:
I will test some applications with these changes. I'll post if find anything.

Also, I was thinking of merging this after a RC2 release? I sent an email to discuss this.

werpu:
Sounds absolutely great.
Thanks for reviewing and testing it.

werpu:
Will take care of the checkstyle violation tomorrow,  I merged the latest code in from master.

werpu:
The pull request has been "stained" by merges from main, I will reopen it with a new number, so that people will get a cleaner diff. Also the internal faces_ts codebase has been updated yesterday to fix a bug regarding script src tags. So it makes sense to issue another pull request.

werpu:
Pull request now back in sync with latest main

melloware:
@werpu can you verify this works in your code: https://github.com/apache/myfaces/pull/342

werpu:
Working on it, there was an issue in the new codebase as well.
The updated code should be in the codebase tomorrow.

PR #327, opened 2022-10-01
MYFACES-4445: websockets refactoring

milansie:
- reworked channel token generation - new token is generated only if needed (ie new session/view), just one token for application scope
- fixed usage of user-specified channel - now also works even in application scope calls
- changed default max idle timeout for websockets - according to omnifaces/mojarra spec
- specified default websocket scope - according to documentation (session scope for user-specified websocket as default)


tandraschko:
can anyone review, too?
it looks good from a first view but i dont have enough time to test and do a deep review

tandraschko:
also we need to port the changes then to 4.0 @milansie, if everyone is ok with it

melloware (approved):

melloware:
These look good to me I am going to merge this and @milansie you can work on the 4.0 changes

PR #338, opened 2022-10-12
MYFACES-4477 Fix Programmatic Views

volosied:
https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4477

Basic test added


impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlTextRendererBase.java
```diff
@@ -143,7 +143,7 @@ protected void renderOutput(FacesContext facesContext, UIComponent component) th
         else
         {
             //default is to escape
-            escape = AttributeUtils.getBooleanAttribute(component, ComponentAttrs.ESCAPE_ATTR, true);
+            escape = AttributeUtils.getBooleanAttribute(component, ComponentAttrs.ESCAPE_ATTR, false);
```
volosied, line 146: Update comment

impl/src/main/java/org/apache/myfaces/view/facelets/impl/DefaultFaceletFactory.java
```diff
@@ -438,6 +438,22 @@ private DefaultFacelet _createCompositeComponentMetadataFacelet(URL url)
...
+                BeanManager bm = CDIUtils.getBeanManager(facesContext);
+                facelet = CDIUtils.get(bm, Facelet.class, true, View.Literal.of(uri));
+            }
+            _managedFacelet.put(uri, facelet != null);
+            if (facelet != null)
+            {
+                return facelet;
+            }
+        }
+
```
volosied, line 456: Code from Thomas' initial fix for this

impl/src/main/java/org/apache/myfaces/application/ViewIdSupport.java
```diff
@@ -482,14 +482,7 @@ protected String handleSuffixMapping(FacesContext context, String requestViewId)
...
 
-        // Otherwise, if a physical resource exists with the name requestViewId let that value be viewId.
-        if (isViewExistent(context,requestViewId))
-        {
-            return requestViewId;
-        }
-        
-        //Otherwise return null.
-        return null;
+        return requestViewId;
```
volosied, line 485: Physical page doesn't exist, so it doesn't make sense to check if the view exists.  Hope this doesn't cause any errors, but the build was successful. I'll see what the TCK says.

melloware (approved):
LGTM but definitely would want @tandraschko to review this.

volosied:
@tandraschko  I've added a test to verify the programmatic facelet works.  TCK also passes with these change. If no other concerns, please merge this.

tandraschko:
yeah but you removed the "view exists" check, which is there by design. So we should run all integration tests to get sure, it still works. we have many viewid-mapping integration tests (not sure if they run currently because of jakarta <> arquillian?)

volosied:
I'm not sure how to run the arquillian tests ( are there any?) 

However, I've refactored this change to keep the "view exists" check.   I now check for the facelet via the CDIUtils in the isViewExistent method. 

Let me know what you think.

impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlTextRendererBase.java
```diff
@@ -142,8 +142,8 @@ protected void renderOutput(FacesContext facesContext, UIComponent component) th
         }
         else
         {
-            //default is to escape
-            escape = AttributeUtils.getBooleanAttribute(component, ComponentAttrs.ESCAPE_ATTR, true);
+            //default is NOT to escape (Changed true -> false, Myfaces-4477)
+            escape = AttributeUtils.getBooleanAttribute(component, ComponentAttrs.ESCAPE_ATTR, false);
```
tandraschko, line 146: isnt this a bug security flaw?

volosied: I've removed the changes and updated the test to not escape the <html> elements. 

I will bring this escaping issue up to the TCK.

PR #342, opened 2022-10-13
MYFACES-4479: The jsf.js script does not read the nonce correctly in …

sidvi1:
MYFACES-4479: The jsf.js script does not read the nonce correctly in modern browsers.

In Chrome it is no longer possible to get a nonce with getAttribute("nonce").
You can only use HTMLElement.nonce (see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/nonce)


werpu:
Closing it, already implemented differently

PR #366, opened 2022-10-30
Fixing localized composite behaviour (3.0.x)

cristof:
Fixes MYFACES-4491

### DIFF
--- a/api/src/main/java/jakarta/faces/component/UIComponent.java
+++ b/api/src/main/java/jakarta/faces/component/UIComponent.java
@@ -1385,6 +1385,52 @@ public abstract class UIComponent
         return paramValue;
     }
 
+    private Resource getLocalizedCompositeResource(String resourceName, String libraryName, FacesContext context)
+    {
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/faces/component/UIComponent.java
```diff
@@ -1842,4 +1841,41 @@ public void setTransient(boolean newTransientValue)
...
     }
+
+    private Resource getLocalizedCompositeResource(String resourceName, String libraryName, FacesContext context) 
+    {
+    	List<String> localizedPaths = getLocalizedPropertiesPaths(resourceName, context);
+    	Resource resource = null;
+    	
+    	for (String path_: localizedPaths) {
+    		resource = context.getApplication().getResourceHandler().createResource(path_, libraryName);
+    		if (resource != null)
```
tandraschko, line 1852: please dont use tabs and make { in new line
no idea, i thought our checkstyle would fail
also better name path_ into localizedPath

cristof: Sorry about that. Fixed style in all four pull requests.

PR #371, opened 2022-11-01
MYFACES-4493: Allow Jakarta and JCP namespaces for FacesComponent

volosied:
I've also changed componentTagDeclarations / unmodifiableComponentTagDeclarations from a  Map to a List to avoid key conflicts and I noticed keys weren't even used (only the values).  Unless a Map was intentional to avoid duplicates, but a set should have been used instead? Nevertheless, Map changed to List and all our unit tests still pass. 

https://issues.apache.org/jira/browse/MYFACES-4493

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/element/FacesConfig.java
+++ b/impl/src/main/java/org/apache/myfaces/config/element/FacesConfig.java
@@ -90,12 +90,12 @@ public abstract class FacesConfig implements Serializable
     }
     
     /**
-     * @since 2.2.0
+     * @since 4.0.0
      * @return 
      */
-    public Map<String, ComponentTagDeclaration> getComponentTagDeclarations()
+    public List<ComponentTagDeclaration> getComponentTagDeclarations()
     {
-        return Collections.emptyMap();
+        return Collections.emptyList();
     }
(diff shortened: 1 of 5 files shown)


impl/src/main/java/org/apache/myfaces/config/element/FacesConfig.java
```diff
@@ -93,9 +93,9 @@ public List<String> getProtectedViewsUrlPatternList()
      * @since 2.2.0
      * @return 
      */
-    public Map<String, ComponentTagDeclaration> getComponentTagDeclarations()
+    public List<ComponentTagDeclaration> getComponentTagDeclarations()
     {
-        return Collections.emptyMap();
+        return Collections.emptyList();
```
volosied, line 98: If this is accepted, how should I update the description?  Keep it as `@since 2.2.0` or update to 4.0.0?  Maybe just mention the new return type?

melloware: Good question.  I would say update to 4.0.0?

melloware (approved):

melloware (approved):

PR #381, opened 2022-11-14
MYFACES-4501: Add xml preamble to every ajax response

volosied:
https://issues.apache.org/jira/browse/MYFACES-4501

### DIFF
--- a/api/src/main/java/jakarta/faces/context/PartialResponseWriter.java
+++ b/api/src/main/java/jakarta/faces/context/PartialResponseWriter.java
@@ -127,9 +127,17 @@ public class PartialResponseWriter extends ResponseWriterWrapper
     public void startDocument() throws IOException
     {
         // Faces 2.2 section 2.2.6.1 Render Response Partial Processing
-        // use writePreamble(...)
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/faces/context/PartialResponseWriter.java
```diff
@@ -130,6 +130,8 @@ public void startDocument() throws IOException
         // use writePreamble(...)
         //_wrapped.write("<?xml version=\"1.0\" encoding=\"utf-8\"?>");
```
melloware, line 131: remove this commented out line?

melloware (changes requested):

api/src/main/java/jakarta/faces/context/PartialResponseWriter.java
volosied: Maybe I should also use writePreamble()? 

I'll update the PR.

volosied: Need to make one more change

melloware (approved):

PR #384, opened 2022-11-15
MYFACES-4501 follow up: Remove writePreamble

volosied:
Since it's now done in the startDocument call.
 
 This caused the preamble to be written twice which in turn caused some pages to fail since the JS wasn't expecting two preambles.


volosied:
I have some additional testing running now to verify no other regressions occur.

melloware (approved):

PR #385, opened 2022-11-16
MYFACES-4425: Save ui input state in repeat

volosied:
https://issues.apache.org/jira/browse/MYFACES-4425

Thoma's fix (in jira comment) might be better, but I opted to just to change the one return for now.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
+++ b/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
@@ -787,6 +787,10 @@ public class MyfacesConfig
     public static final String EL_RESOLVER_TRACING = "org.apache.myfaces.EL_RESOLVER_TRACING";
     public static final boolean EL_RESOLVER_TRACING_DEFAULT = false;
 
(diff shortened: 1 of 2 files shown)


melloware (approved):

tandraschko:
IMO we should not make it per default
we didnt had any user since 10 years which requires this

can someone start a dicussion with mojarra team?

otherwise i would introduce a webcontext param

volosied:
@tandraschko Following up regarding the comment above.

volosied:
I added a new web parameter: `org.apache.myfaces.UI_REPEAT_STATE_SAVING`, but the new behavior is the default.

pnicolucci:
@tandraschko are you ok with this enabled by default? Or are you thinking we should create a TCK Challenge for this test/behavior? I think we need to either make this the default behavior or challenge the tck.

tandraschko:
Could anyone check that case again and talk with the Mojarra guys, if this case is really required?
As i said in my small statement in the linked issue, we have this "issue" since 10 years propably and noone ever openend a JIRA issue
maybe because the case is not that valid

Otherwise i'm ok with it

volosied:
Challenge was accepted, so I'll close this PR.

PR #396, opened 2022-11-25
MYFACES-4509: Quarkus remove view init optimization

melloware:


tandraschko:
@melloware thats just not enough
we need to remove all configs as they are NOT overwriteable by a user
also the primefaces settings

melloware:
OK i will remove them all!

PR #399, opened 2022-11-28
MYFACES-4503: use ELManager.getExpressionFactory()

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4503


tandraschko:
should we leave the old lookup as fallback if ELManager returns null?

pnicolucci:
> should we leave the old lookup as fallback if ELManager returns null?

@tandraschko I'd prefer not to unless we find that it is necessary this keeps the code clean IMO. Looking at the ELManager Java Doc: https://jakarta.ee/specifications/expression-language/5.0/apidocs/jakarta.el/jakarta/el/elmanager#getExpressionFactory() and the Tomcat implementation for example https://github.com/apache/tomcat/blob/main/java/jakarta/el/Util.java#L97 I think we'll always get an instance back.

tandraschko:
all right, then +1

PR #414, opened 2022-12-09
2.3 next: https://issues.apache.org/jira/browse/MYFACES-4526

werpu:


werpu:
Note, if you still need the i18n, use 
        <h:outputScript pr:nonce="test123"  name="faces-i18n.js" library="myfaces" target="head"/>

PR #224, opened 2021-09-19
MYFACES-4412: Add new layout="list"

melloware:
MYFACES-4412: Add new layout="list" attribute value to h:selectManyCheckbox and h:selectOneRadio so it renders `<ul>` instead of `<table>`

Also the `"border"` attribute was being completely ignored and always set to "0".  Followed Mojarra's example for this fix.

See: https://github.com/eclipse-ee4j/faces-api/issues/1574

PR #419, opened 2022-12-09
[SPEC 1574 / MYFACES-4528 ] Ensure top level ul does not have border attribute

KyleAure:
Based on the SPEC issue here: https://github.com/jakartaee/faces/issues/1574

When layout="list" the border attribute should be ignored. 

When a form like this:
```xml
<h:form id="form">
	<h:selectManyCheckbox id="input" layout="list" border="5">
          ...
	</h:selectManyCheckbox>
</h:form>
```

Is rendered the current output is: 
```xml
<html>
  <head/>
  <body>
    <form id="form" name="form" method="post" action="/LayoutAttribute/selectManyCheckBox.xhtml" enctype="application/x-www-form-urlencoded">
      <ul id="form:input" border="5">
        ...
      </ul>
  </body>
</html>
```

Based on the spec we should be rendering: 
```xml
<html>
  <head/>
  <body>
    <form id="form" name="form" method="post" action="/LayoutAttribute/selectManyCheckBox.xhtml" enctype="application/x-www-form-urlencoded">
      <ul id="form:input">
        ...
      </ul>
  </body>
</html>
```

This is because we are incorrectly starting this element with table passthrough properties, instead of ul passthrough properties.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlCheckboxRendererBase.java
+++ b/impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlCheckboxRendererBase.java
@@ -141,7 +141,14 @@ public class HtmlCheckboxRendererBase extends HtmlRenderer
         ResponseWriter writer = facesContext.getResponseWriter();
 
         writer.startElement(usingTable != null ? HTML.TABLE_ELEM : HTML.UL_ELEM, selectMany);
-        HtmlRendererUtils.renderHTMLAttributes(writer, selectMany, HTML.SELECT_TABLE_PASSTHROUGH_ATTRIBUTES);
+        if(usingTable != null) 
+        {
+            HtmlRendererUtils.renderHTMLAttributes(writer, selectMany, HTML.SELECT_TABLE_PASSTHROUGH_ATTRIBUTES);
+        }
+        else 
+        {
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
These look good to me thanks for the PR.  They might want you to open a JIRA ticket if you have not already so we can track this fix.

KyleAure:
@melloware Thanks for the heads up.  I created the issue here: https://issues.apache.org/jira/browse/MYFACES-4528
Sorry, should have done that first.  This is my first commit to this community and didn't realize you were using Jira

volosied:
This looks good -- just tested locally. Thanks for providing a patch and making the JIRA.

Related to which, I think the community should create an issues tab to allow users to create issues in this repo since JIRA is disabled for new signups.  We can follow on on this later.

melloware:
I actually prefer GitHub Issues myself but yes an easy way for users to report issue we can always make a JIRA ticket if we discover its a real issue

melloware:
Also does this fix need to be backported to 2.3 and 2.3-next?

volosied:
@melloware  It looks like the layout=list was only added to 4.0 (per the spec). I thought it was added to 2.3-next, but was mistaken. So no need to backport.

https://github.com/apache/myfaces/pull/224

PR #426, opened 2022-12-12
https://issues.apache.org/jira/browse/MYFACES-4040: 

werpu:
Fix for MYFACES-4040, backport to the official 2.3 codebase @pnicolucci should decide whether we take it in (or the entire community)


werpu:
Just a short explanation. This is not a full backport of the entire next javascript rendering infrastructure, just a small fix which shifts the parameters around just like the other fixes do, but within the code of 2.3.x:
Example:
```html
<a href="#" onclick="jsf.util.chain(this, event,'jsf.ajax.request(this,event,{execute:\'submitButton \',render:\'output \',params:{\'Hello World\':\'Hello from f:param2\',\'hello2\':\'Hello from f:param\',\'javax.faces.behavior.event\':\'action\'}})'); return false;" id="submitButton" name="submitButton">Click here</a>
````
If you check the params are now passe dunder the params attribute. and so is the javax.faces.behavior.event  (which is a request parameter)

werpu:
Checking why the build is failing

werpu:
Tests have passed... I will leave it up to you whether you want to take it in, but given we have synchronized our myfaces jsf.js codebases now back from next into the original codebase, I highly can recommend to take this also in and cosider this bug fixed for all of 2.3+

pnicolucci:
@werpu @volosied  I don't have an objection to backporting this one.

werpu:
Ok I will merge the code then into 2.3, thanks for the update.
Also @pnicolucci I have merged the jsf.js 2.3-next codebase back into 2.3, this however means we cut browser support for browsers below IE9.
If there are any objections to this - I know you guys have a ton of customers, so a change in a stable branch might be too much, I will revert this change.

The codebase is more or less the same but it has dropped a ton of ie6 and 7 and even older browser code, and introduced a construct in a fixed head replace (there was a tck issue this has fixed) which relies on a dom construct which ie9+ only support.

pnicolucci:
> Ok I will merge the code then into 2.3, thanks for the update. Also @pnicolucci I have merged the jsf.js 2.3-next codebase back into 2.3, this however means we cut browser support for browsers below IE9. If there are any objections to this - I know you guys have a ton of customers, so a change in a stable branch might be too much, I will revert this change.
> 
> The codebase is more or less the same but it has dropped a ton of ie6 and 7 and even older browser code, and introduced a construct in a fixed head replace (there was a tck issue this has fixed) which relies on a dom construct which ie9+ only support.

@werpu I think this requires more of a broad community discussion on the mailing list. The last I saw I thought the decision was to only update the 4.0 (potentially 2.3-next) code base and to potentially backport to previously released versions with a context parameter. I'd like to see a vote so we can ensure everyone is on board with the decision, pros/cons, and default of any such context parameter. I'm fine with backporting fixes, however, removing support for browsers etc IMO warrants a more detailed discussion. My personal opinion is we err on the side of caution.

werpu:
I have started an informal vote on the mailing list to get a general community opinion about it.

werpu:
I found a better way.  A small code reorg into existing structures, allows me to get a cleaner cut between legacy browser support and modern browsers. Basically I just reorganized smaller parts of the code into the already existing shim layer.
I am now able to sync more or less with minor code changes between the branches.
This already is merged, I will reissue the fix for 4040 later.

werpu:
works with all the latest fixes in 2.3! I will merge now this backend pull request, given it is uncritical.
Will also port this tomorrow to 3.0!

PR #435, opened 2022-12-16
Fix MYFACES-4533 - prefix id in name attribute

volosied:
https://issues.apache.org/jira/browse/MYFACES-4533

### DIFF
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxResponse.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxResponse.js
@@ -126,7 +126,7 @@ _MF_SINGLTN(_PFX_XHR + "_AjaxResponse", _MF_OBJECT, /** @lends myfaces._impl.xhr
              * definitely not a null value to avoid type confusions later on
              */
             mfInternal.namingModeId = (partials.id || "");
-            mfInternal.namespaced =  (partials.namedViewRoot || "");
+            mfInternal.namespaced =  (partials.attributes.getAttribute("namedViewRoot") || ""); // getAttribute not a function?
 
             var childNodesLength = partials.childNodes.length;
 
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxResponse.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxResponse.js
@@ -126,7 +126,7 @@ _MF_SINGLTN(_PFX_XHR + "_AjaxResponse", _MF_OBJECT, /** @lends myfaces._impl.xhr
              * definitely not a null value to avoid type confusions later on
              */
             mfInternal.namingModeId = (partials.id || "");
-
+            mfInternal.namespaced =  (partials.namedViewRoot || "");
 
             var childNodesLength = partials.childNodes.length;
 
@@ -197,6 +197,7 @@ _MF_SINGLTN(_PFX_XHR + "_AjaxResponse", _MF_OBJECT, /** @lends myfaces._impl.xhr
          * faces 2.3 we set all the viewstates under a given declared viewRoot or all forms
          * if none is given
          */
+
         this._updateFacesClientArtifacts(context,  mfInternal.appliedViewState, this.P_VIEWSTATE);
     },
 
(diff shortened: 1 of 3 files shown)


melloware (approved):

volosied:
Javascript needs updates here... 

https://github.com/apache/myfaces/blob/8a220ce7f7e8a8fdb1d024a3ebddac0b3899be5d/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxResponse.js#L261

Need to obtain the namespace container prefix

melloware:
Not sure if this matters but mojarra does this in PartialResponseWriter and it doesn't look like MyFaces does?

It looks like it is doing NamingContainer View ID code just thought I would mention it.

```java
@Override
    public void startDocument() throws IOException {
        ResponseWriter writer = getWrapped();
        String encoding = writer.getCharacterEncoding( );
        if( encoding == null ) {
            encoding = "utf-8";
        }
        writer.writePreamble("<?xml version='1.0' encoding='" + encoding + "'?>\n");
        writer.startElement("partial-response", null);
        FacesContext ctx = FacesContext.getCurrentInstance();
        if (null != ctx && ctx.getViewRoot() instanceof NamingContainer) {
            String id = ctx.getViewRoot().getContainerClientId(ctx);
            writer.writeAttribute("id", id, "id");
        }
    }
```

volosied:
Thanks for pointing that out. I noticed that we always write the id in the `"partial-response` tag, but Mojarra only writes it under the condition you pointed above (viewroot instanceof NamingContainer).

However, this leads to a problem because Mojarra seems to be out of compliance here per the schema here: https://jakarta.ee/xml/ns/jakartaee/web-partialresponse_4_0.xsd 

```
<xsd:element name=“partial-response” type=“jakartaee:partial-responseType”>
<xsd:annotation>
<xsd:documentation>
[ <p> The “partial-response” element is the root of the partial response information hierarchy, and contains nested elements for all possible elements that can exist in the response.</p> <p>This element must have an “id” attribute whose value is the return from calling getContainerClientId() on the UIViewRoot to which this response pertains. </p>
</xsd:documentation>
</xsd:annotation>
</xsd:element> 
```

_"This element must have an “id” attribute whose value is the return from calling getContainerClientId() on the UIViewRoot to which this response pertains. "_  Mojarra should write it every time, not just in some cases.

The question becomes how can view know if the viewroot is an  instance of NamingContainer in the ajax responses. 

This is as far I've looked into it and will try to get a conversation started in our existing head id challenge. 

Links:
MyFaces
https://github.com/volosied/myfaces/blob/3984a6dfb7960f2ad6738653f2beea55f7ce154f/impl/src/main/java/org/apache/myfaces/context/servlet/PartialViewContextImpl.java#L437

Mojarra: 
https://github.com/eclipse-ee4j/mojarra/blob/baf7998ef3492742d7e407c48c2fc3e87cb758c7/impl/src/main/java/com/sun/faces/context/PartialViewContextImpl.java#L437

https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-3973?filter=allissues

melloware:
Good question.  I think we should stay in sync with whatever Mojarra is doing for consistency if that helps?

tandraschko:
i dont understand this for 100%
we now add the namespace to the name attr but not id?

volosied:
Currently (RC2), we don't. We only render `name="jakarta.faces.ViewState"`.  This PR prefixes with the id (via Java) now so it looks more like: `name="j_id__v_0:jakarta.faces.ViewState"`. 

However, it doesn't address the cases where the input is rendered via JavaScript.  After an Ajax response, the updated form has  `name="jakarta.faces.ViewState"` again.

volosied:
@werpu  any chance you could look at this?

werpu:
Sorry for jumping in so late, I have not had that on my radar...

Determine the naming container id, I do that with the prefix passed via the identifier.

The JSDOC spec says:

image

Thats basically what we implement in our javascripts.

So in order to prefix whatever we have.. we just use the same name as the id...

On the other hand.. in the submit request the viewstate is always passed down only as javax.faces.ViewState as per spec

as per JSDoc:
image

So it makes zero difference whether we have the viewrootid in the name or not (names are unique only per form, and we only have one viewstate element per form, identifiers must be unique per page), but it would break the spec passing the viewroot id in the viewstate!

I would not prefix the name frankly spoken and if you need it what is the use case for it? If there is a use case we have to challenge the spec on this aspect, because it clearly wants a non prefixed name on this.

PR #443, opened 2022-12-19
fix: enable integration tests

bohmber:
There are two problems with the integration tests.

- The Maven embedded build in the integration tests can't resolve the produced artifacts of myfaces because that is a new Maven build. Moving the execution of the integration tests after the package phase allows to reference the produced war with the myfaces artifacts.
- chrome should be headless


melloware (approved):

melloware:
Excellent work!

volosied:
Thank you!

PR #459, opened 2022-12-24
Quarkus 3 support

melloware:
@tandraschko  I got it all working!

Quarkus did not like the ViewScopedEventListener.Bridge being an inner scoped Application scoped static bean. I moved it into its own normal class and then Quarkus was able to discover it!

### DIFF
--- a/extensions/quarkus/showcase/src/test/java/org/apache/myfaces/core/extensions/quarkus/showcase/QuarkusMyFacesShowcaseTest.java
+++ b/extensions/quarkus/showcase/src/test/java/org/apache/myfaces/core/extensions/quarkus/showcase/QuarkusMyFacesShowcaseTest.java
@@ -69,7 +69,7 @@ public class QuarkusMyFacesShowcaseTest {
         final HtmlPage page = webClient.getPage(url + "/index.xhtml");
         final HtmlDivision datatable = (HtmlDivision) page.getElementById("form:carTable");
         assertThat(datatable).isNotNull();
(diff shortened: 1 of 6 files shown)


extensions/quarkus/showcase/src/test/java/org/apache/myfaces/core/extensions/quarkus/showcase/QuarkusMyFacesShowcaseTest.java
```diff
@@ -69,7 +69,7 @@ public void shouldOpenIndexPage() throws Exception {
         final HtmlPage page = webClient.getPage(url + "/index.xhtml");
         final HtmlDivision datatable = (HtmlDivision) page.getElementById("form:carTable");
         assertThat(datatable).isNotNull();
-        assertThat(datatable.getByXPath("//tr[contains(@role,'row') and contains(@class,'ui-datatable-selectable')]"))
+        assertThat(datatable.getByXPath("//tr[contains(@class,'ui-datatable-selectable')]"))
```
melloware, line 72: PF 12 removed the role="row" as not needed.

melloware:
@bohmber I dont understand the Apache RAT build failure?

bohmber:
> @bohmber I dont understand the Apache RAT build failure?

@melloware Most of the time apache rat plugin is not configured in the pom or configured via parent poms.
I think it's bad practice to declare a pom without a parent in a multimodule project and declare a different version. That is the case in the showcase project.

melloware:
got it. let me fix

PR #462, opened 2022-12-24
build: quarkus extension showcase groupId should be a myfaces groupid…

bohmber:
… and logs should be created in target directory


melloware:
Thanks!

tandraschko:
Should we also skip the showcase for deployment, like we did for integration Tests?

melloware:
AHhh yes we should

PR #471, opened 2023-01-05
MYFACES-4541: NPE check on destroying view

melloware:


tandraschko:
Yes
In past ViewScope SPI was replaceable but we only support CDI beans and scopes now, so its not required anymore

PR #473, opened 2023-01-10
MYFACES-4543: Use Flash Cookies When In Client Window URL Default Mode

volosied:
https://issues.apache.org/jira/browse/MYFACES-4543

Code checks for FacesConfigurator.isEnableDefaultWindowMode which is when url is is enabled by default (i.e. not user specified):

https://github.com/apache/myfaces/blob/fcec6d6dc9bd13de0c3f728215bc02e9237316c8/impl/src/main/java/org/apache/myfaces/config/FacesConfigurator.java#L1453-L1482

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/context/flash/FlashImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/context/flash/FlashImpl.java
@@ -19,6 +19,7 @@
 package org.apache.myfaces.context.flash;
 
 import org.apache.myfaces.util.lang.SubKeyMap;
+import org.apache.myfaces.config.FacesConfigurator;
 import org.apache.myfaces.config.webparameters.MyfacesConfig;
 import org.apache.myfaces.util.ExternalContextUtils;
 
@@ -128,6 +129,7 @@ public class FlashImpl extends Flash implements ReleasableFlash
     
      // ~ static methods  -----------------------------------------------------
     
+    private boolean _defaultWindowModeEnabled = false; // MYFACES-4543
     /**
      * Return a Flash instance from the application map
      * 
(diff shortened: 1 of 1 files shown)


tandraschko:
is this really correct?
wouldnt it be better to check if there currently is a ClientWindow available for the current request?

volosied:
We should have some clean up? I don't understand why we have `org.apache.myfaces.DEFAULT_WINDOW_MODE` (default url) and `jakarta.faces.CLIENT_WINDOW_MODE` (default none)

https://github.com/apache/myfaces/blob/89c747e85615e3f33265e664c8361789f38ea7db/impl/src/main/java/org/apache/myfaces/lifecycle/clientwindow/ClientWindowFactoryImpl.java#L37

tandraschko:
Yeah, seems so

volosied:
At the same time, if we use cookies if url mode is enabled by the user then why not just remove the client window checks here? 


I think I'd prefer to leave the PR as is, but we could revisit later on if needed.

tandraschko:
i think we need to dig a bit deeper
i understand your problem as URL ClientWindow doesnt do something like a initialRedirect, so when you visit the view and press f5, the clientWindow is will be regenerated

AFAICS Mojarra doesnt even use the ClientWindow for flash stuff
why did MyFaces introduce this?

volosied:
Yes, I saw Mojarra used cookies for flash too (at least for the scenario in the TCK). No idea why MyFaces changed. 

I see some mention here: https://issues.apache.org/jira/browse/MYFACES-3404 (`Add windowId to server side state saving algorithm`) 

```
 There are two points where this logic could be useful inside MyFaces Core :

1. Server side state
2. Flash scope

But in practice the only really relevant is 1. To implement 2. it is necessary
a "requestId". MyFaces FlashImpl uses a counter and store its value inside
a cookie. To fix this scope properly, the best is create a ExternalContext
wrapper and provide and alternate Flash object, but that sounds like something
that can be done outside MyFaces Core. If you are using CDI scopes, it sounds
reasonable to provide an alternate Flash scope in CODI.
```

volosied:
@tandraschko Would you prefer to use flash cookies altogether and move away from client window logic?

tandraschko:
I would do the following:
1) change your IF to "isUrlMode", also when the user sets it
2) please add a comment and link to the issue number

PR #492, opened 2023-01-20
MYFACES-4551: Allow group validation for uiselectone elements

volosied:
https://issues.apache.org/jira/browse/MYFACES-4551

Various scenarios tested in the TCK, see: https://github.com/jakartaee/faces/blob/c6d7d879065056978d74e5f486847fb203606203/tck/faces23/uiinput/src/main/webapp/spec329.xhtml 

This TCK test has been a pain to debug in our impl and to understand. I've added comments and concerns for clarity. Thanks.

### DIFF
--- a/api/src/main/java/jakarta/faces/component/UISelectOne.java
+++ b/api/src/main/java/jakarta/faces/component/UISelectOne.java
@@ -55,7 +55,7 @@ public class UISelectOne extends UIInput
 
     public static final String INVALID_MESSAGE_ID = "jakarta.faces.component.UISelectOne.INVALID";
 
-    private boolean selectItemValueFound = false;
+    private boolean previouslySubmittedOrValidated = false;
 
     public UISelectOne()
     {
@@ -69,9 +69,10 @@ public class UISelectOne extends UIInput
     }
 
     /**
-     * Verify that when ever there is a ValueExpression and submitted value is not empty, then
+     * Check whether a group exists and then
      * visit all the UISelectItem elements within the UISelectOne radio components to check if
-     * the submitted value exists in any of the select items.
+     * the submitted value is empty (ie. not submitted) or if a previous group item has been
+     * has failed to be validated (if no so further validation processing is needed)
      *
      * @see jakarta.faces.component.UIInput#processValidators(jakarta.faces.context.FacesContext)
      */
@@ -79,9 +80,9 @@ public class UISelectOne extends UIInput
     public void processValidators(FacesContext context) 
     {
         String group = getGroup();
-        ValueExpression ve = getValueExpression("value");
         String submittedValue = (String) getSubmittedValue();
-        if (group != null && !group.isEmpty() && ve != null && !isEmpty(submittedValue)) 
+
+        if (group != null && !group.isEmpty()) 
         {
             final UIComponent form = getRadioNestingForm(context, this);
 
@@ -90,32 +91,25 @@ public class UISelectOne extends UIInput
                 @Override
                 public VisitResult visit(VisitContext visitContext, UIComponent target) 
                 {
+                    // check they they are of the same group
                     if (target instanceof UISelectOne  && ((UISelectOne) target).getGroup().equals(group)) 
                     {
-                        UISelectOne radio = (UISelectOne) target;
-
-                        // if target is an instance of UISelectOne then get all the UISelectItem children
-                        // and verify if the submitted value exists
-                        for (Iterator<UIComponent> iter = radio.getChildren().iterator(); iter.hasNext(); ) 
-                        {
-                            UIComponent component = iter.next();
-                            if (component instanceof UISelectItem) 
-                            {
-                                UISelectItem item = (UISelectItem) component;
-                                if (item.getItemValue().equals(submittedValue)) 
-                                {
-                                    selectItemValueFound = true;
-                                    return VisitResult.COMPLETE;
-                                }
-                            }
-
-                        }
-                        return VisitResult.REJECT;
+                        // check if the is empty (see ) or if it's not valid (means this path has been taken already)
+                        if(isEmpty(submittedValue) && (( (EditableValueHolder)target).isLocalValueSet() || !((EditableValueHolder)target).isValid() )){
+                            previouslySubmittedOrValidated = true;
+                            return VisitResult.COMPLETE;
+                        }          
                     }
-
                     return VisitResult.ACCEPT;
                 }
             });
+        }
+
+        if(previouslySubmittedOrValidated){
+            // Skip further validation due to either 
+            // 1) one of the submissions are not valid (for instance, required, but none submitted (see jakarta/faces#329))
+            // 2) submitted value has been found and validated
+            return;
         }
         
         super.processValidators(context);
@@ -186,9 +180,9 @@ public class UISelectOne extends UIInput
             }
         }
 
-        // if selectItemValueFound is true, then it means that we have found the select item value
+        // if previouslySubmittedOrValidated is true, then it means that we have found the select item value
         // in the other UISelectOne radio components, and no validation error is thrown.
-        if (selectItemValueFound) 
+        if (previouslySubmittedOrValidated) 
         {
             return;
         }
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/apache/myfaces/renderkit/RendererUtils.java
```diff
@@ -693,17 +693,25 @@ else if (allowNonArrayOrCollectionValue)
...
     {
-        if (submittedValue != null && !(submittedValue instanceof String))
+
+        if(submittedValue == null)
+        {
+            if (log.isLoggable(Level.FINE))
+            {
+                log.fine("No conversion necessary for null uiselectone value: client id " + output.getClientId());
+            }
+            return null;
```
volosied, line 703: Mojarra returns null and so I did the same here ( since no value is submitted)

I'm just not sure if this is correct, however.

volosied:
Mojarra's steps: 


1) MenuRenderer#decodeGroup -- submitted value is set to RIConstants.NO_VALUE ("") - https://github.com/eclipse-ee4j/mojarra/blob/3fd5d16cc3c0c631c97a0a62e8e68a6be66fbafe/impl/src/main/java/com/sun/faces/renderkit/html_basic/MenuRenderer.java#L872
2) UIInput#validate- submittedValue is not not null, so it continues with validation. The if checks for ((String) submittedValue).length() == 0 and sets the submitted value back to null -  https://github.com/eclipse-ee4j/mojarra/blob/be054fde83df012a6add3b406eec44aa28975a28/impl/src/main/java/jakarta/faces/component/UIInput.java#L806-L811
3) MenuRenderer#convertSelectOneValue - during validate() a getConvertedValue call occurs, which leads to MenuRenderer#convertSelectOneValue. In this method, there's a isNoValueOrNull call which sees that the submitted value is null and logs ""No conversion necessary for SelectOne Component since the new value is null"

These changes followed a similar pattern.   Although, I believe parts of this could be a potential challenge, such as the null check in the conversion call.

impl/src/main/java/org/apache/myfaces/renderkit/html/util/HtmlRendererUtils.java
```diff
@@ -292,6 +292,12 @@ public static void decodeUISelectOne(FacesContext facesContext, UIComponent comp
                                 callback);
                     }
                 }
+                else 
+                {
+                    // means input was not submitted. set to empty string so we can validate required fields
+                    // if not set, a null value will skip validation -- see beginning of UIInput#validate
+                    ((EditableValueHolder)component).setSubmittedValue(RendererUtils.EMPTY_STRING);
```
volosied, line 299: Setting to RendererUtils.EMPTY_STRING, similar as to Line 322 & 245

Not a fan of using an empty string here ( since it's latter converted to a null), maybe an enum would be better...?

impl/src/main/java/org/apache/myfaces/renderkit/RendererUtils.java
volosied: Might be worth a challenge? Unless we switch to enums which would signify that no value is inputted, hence no conversion necessary (rather than rely on this empty string / null nonsense).

api/src/main/java/jakarta/faces/component/UISelectOne.java
```diff
@@ -69,19 +69,20 @@ public String getFamily()
...
      */
     @Override
     public void processValidators(FacesContext context) 
     {
         String group = getGroup();
-        ValueExpression ve = getValueExpression("value");
         String submittedValue = (String) getSubmittedValue();
-        if (group != null && !group.isEmpty() && ve != null && !isEmpty(submittedValue)) 
+
+        if (group != null && !group.isEmpty())
```
volosied, line 85: Relevant spec: https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/uiselectone#processValidators(jakarta.faces.context.FacesContext)

If [getGroup()](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/uiselectone#getGroup()) is set, and [UIInput.getSubmittedValue()](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/UIInput.html#getSubmittedValue()) is empty, and at least one other component having the same group within a UIForm parent has a non-empty [UIInput.getSubmittedValue()](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/UIInput.html#getSubmittedValue()) or returns true on [UIInput.isLocalValueSet()](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/UIInput.html#isLocalValueSet()) or returns false on [UIInput.isValid()](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/UIInput.html#isValid()), then skip validation for current component, else perform standard superclass processing by super.processValidators(context).

volosied: I need to check for the "non-empty UIInput.getSubmittedValue()" case.

impl/src/main/java/org/apache/myfaces/renderkit/RendererUtils.java
melloware: Might be a good question in a challenge for BalusC and Arjan to answer?

melloware (approved):
Overall this refactoring looks pretty good to me.

volosied:
I'd also like to merge this for RC4, unless there are objections? 

Any follow up / improvements can be done in a future RC release. 

@melloware  @tandraschko

melloware:
Same as before good with me.  we can always fix in RC5 :)

tandraschko:
I will review tomorrow

impl/src/main/java/org/apache/myfaces/renderkit/html/util/HtmlRendererUtils.java
volosied: Not sure if enums are spec allowed due to this isEmpty() definition: 
https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/uiinput#isEmpty(java.lang.Object)
```
If the value is null, return true. If the value is a String and it is the empty string, return true. If the value is an array and the array length is 0, return true. If the value is a List and the List is empty, return true. If the value is a Collection and the Collection is empty, return true. If the value is a Map and the Map is empty, return true. In all other cases, return false. 
```

I think we will need to keep it as is (set to empty string when no value submitted, if kept as null it would cause validation to be skipped)

api/src/main/java/jakarta/faces/component/UISelectOne.java
```diff
@@ -90,37 +91,43 @@ public void processValidators(FacesContext context)
...
-                                UISelectItem item = (UISelectItem) component;
-                                if (item.getItemValue().equals(submittedValue)) 
-                                {
-                                    selectItemValueFound = true;
-                                    return VisitResult.COMPLETE;
-                                }
-                            }
-
-                        }
-                        return VisitResult.REJECT;
```
tandraschko, line 113: @volosied this code checked if the submittedValue is valid and present as SelectItem - do we still have such a check now after removing this?

volosied: My understanding is that submittedValue is checked via UISelectOne#validateValue which calls into: https://github.com/apache/myfaces/blob/89c747e85615e3f33265e664c8361789f38ea7db/api/src/main/java/org/apache/myfaces/core/api/shared/SelectItemsUtil.java#L177

Additionally, the spec does not ask for the `item.getItemValue().equals(submittedValue)` / `component instanceof UISelectItem` checks here. 
https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/component/uiselectone#processValidators(jakarta.faces.context.FacesContext)

volosied: Additionally, the current code causes a failure with the Spec329 test: 


```
Caused by: java.lang.ClassCastException: class java.lang.String cannot be cast to class ee.jakarta.tck.faces.test.javaee8.uiinput.Spec329Entity (java.lang.String is in module java.base of loader 'bootstrap'; ee.jakarta.tck.faces.test.javaee8.uiinput.Spec329Entity is in unnamed module of loader com.ibm.ws.classloading.internal.AppClassLoader @e7d2dc5)
at ee.jakarta.tck.faces.test.javaee8.uiinput.Spec329Entity.equals(Spec329Entity.java:33)
```

The new work works fine for me and if you use submit some other value in the form (other than the available options), you'll just get this type of error.
```inDataTableWithEntityList:table:0:radio: Validation Error: Value is not valid```


As far as I can tell, this code is good.

volosied:
Need to check on more edge case before merging

volosied:
Last commit update -- Empty string / null means there is no value -- do not convert them for ui select one.

volosied:
All checks passed -- merging.

PR #502, opened 2023-01-24
Fixes for TCK 790 on main

werpu:
Fixes for the old codebase for main, server implementation still pending (need to do someone else)

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/renderkit/html/ParamsNamingContainerResolver.java
+++ b/impl/src/main/java/org/apache/myfaces/renderkit/html/ParamsNamingContainerResolver.java
@@ -28,9 +28,9 @@ import jakarta.faces.render.ResponseStateManager;
 import java.util.Map;
 
 /**
- * A centralized wrapper class which deals with naming container resolution
+ * A centralized wrapper class, which deals with naming container resolution
  * on params level.
- * The problem is, the naming container ajax requests introduce per spec
+ * The problem is, the naming container Ajax requests introduce per spec
  * that the parameters need to be prefixed. This works for normal components
  * but not standardized request parameters, we have to prefix them for a params
  * lookup.
@@ -42,6 +42,9 @@ import java.util.Map;
  */
 public class ParamsNamingContainerResolver
 {
+    public static final String CACHE_ATTR_NAMING_PREFIX = "myfaces.cache.namingcontainerprefix";
+    public static final String EMPTY = "";
+    public static final String CACHE_ATTR_POSTBACK = "myfaces.cache.postback";
     Map<String, String> delegate;
     FacesContext facesContext;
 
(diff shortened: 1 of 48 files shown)


werpu:
Do not merge this in util the server side implementation is fixed to be able to deal with namespaces.
You can use it as a base for the namespace tck though

werpu:
closing it for now

werpu:
Reopened after fixing a last minute issue

melloware:
LGTM

melloware:
@werpu sorry asked you to weigh in on a similar issue i think you ran into with HTMLUnit OmniFaces IT tests are wonky with MyFaces 4 but i think its HTMLUnit related.  See: https://github.com/omnifaces/omnifaces/issues/729

werpu:
Answered... yes it very likely is a similar isse, the code affected probably uses valid Ecmascript constructs, which Rhino simply cannot handle anymore!

impl/src/main/java/org/apache/myfaces/renderkit/html/ParamsNamingContainerResolver.java
```diff
@@ -0,0 +1,124 @@
...
+        {
+            Map<String, String> reqParamMap = facesContext.getExternalContext().getRequestParameterMap();
+
+            //no prefix, we have a blank ViewState in the request!
+            if(reqParamMap.containsKey(ResponseStateManager.VIEW_STATE_PARAM))
+            {
+                return "";
+            }
+            // TODO please optimize this code, we probably can store the data below on request level
+            // it is static per request!
```
melloware, line 101: Can this TODO be removed?

werpu:
I left this todo in as point of discussion, see also my last comment on MyFaces 4533!
I will fix it as soon as everyone is ok with a strategy on how we optimize it!
This is not ready for merge yet, I will merge probably tomorrow!

werpu:
Btw. I already have prepared fixes for master, 2.3-next and 2.3 is still in work, so once we merge this code in, expect a fix for master relatively quickly, 2.3-next and 2.3 need some additional testing!

melloware:
Nice work!  I appreciated the persistence!

tandraschko:
TBH im not sure if we should backport it to older branches
viewroot namespacing is only used in portlets and no portlet impl uses myfaces
so this is only for be TCK compatible and will never be used in production for now

also we must optimize the server side stuff
always a new object instance (ParamsNamingContainerResolver) and so on isnt quite good for a feature which isnt used

werpu:
Will check the failing test now...
The standard unit tests all pass result of the TCK tests please see the MYFACES-4533 issue!

werpu:
I can live with not porting it to the older implementations, but for the sanity of keeping the javascript implementations together I need to have this functionality on the client side enabled. (aka take at least the js changes in)

werpu:
Fix comitted, also my 21 local ajax tests have passed now:
image
Spec 790 passes as well:
image

Also the tcks pass as well as before on 2.2 level:
image

the failing test is a server side problem and reported!

volosied:
Thank you for all your hard work on this.  Glad to see it's finally passing. I think we should be at 100% for the TCK with this change now :) 

As for the PR, it looks like it could be rebased since it contains older commits? 

I also created this small commit for caching the prefix if you wanted to use it: https://github.com/volosied/myfaces/commit/22d451c7b7a7f01b16060d21a5563205680b7975

It should keep things in the current faces context until it's released. 

https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/context/facescontext#getAttributes()

werpu:
not sure 100% see the failing 2.2 test regarding the ui repeat and savestates (image above), which is clearly a bug on the implementation side. I will add it the prefix caching.

werpu:
Added the caching!
Tests still pass, TCK as well!

tandraschko:
TBH i dont like the way how ParamsNamingContainerResolver works
this breaks external libs like PrimeFaces

i think e.g. context.getExternalContext().getRequestParameterMap() should be do all the work internally

if you need help to implement this, i can do it for you
please just rebase the PR here and with only JS changes
and provide a second PR with server side changes, so i can take a look in detail

werpu:
Thanks for the suggestion, does it really break the external libs? 

The thing is it basically just prefixes the viewState on the param map lookup if you go in with a standard parameter.

I cannot omit that the request parameters are coming in prefixed for the naming container case that is per spec and thats what we have done wrongly so far!

But if you pull the values in from the parameter map you have to be prefix aware anyway. The resolver just does this internally in a non intrusive way, is the param coming in prefixed it wont prefix, if it comes in non prefix it tries both a prefix and non prefixed lookup!

But either way I will provide a second pull request with the scripts only, so that you can have a look at yourself, and will postpone the merge. Feel free to take over the server side for a better solution!

werpu:
https://github.com/apache/myfaces/pull/510 
Pull request opened @tandraschko feel free to take over!

tandraschko:
AFAIR all params are prefixed, thats how we implemented it with liferay team: https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/resources/META-INF/resources/primefaces/core/core.ajax.js#L607

werpu:
Yes thats more or less what the current js code dows now!
Per spec the sent params now are prefixed with the naming container id!

tandraschko:
perfect
the TCK just adds a overwritten viewRoot in the testproject then?

werpu:
Yes thats pretty much where it fails, the failing Spec790 test issues an implicit ajax navgiation over an action outcome, the root cause simply is that the impl fails to detect the jakarta.faces.ViewState and other parameters properly because internally they were not prefixed, and thats pretty much all my fix and the code I got from @volosied I was basing my fix on, was about!

volosied:
@werpu @tandraschko 

I know Spec790 has has taken lot of our time. I suppose Thomas could try to find a solution for the java code side of it (would be great if you could), but, if not, we could challenge it instead?

tandraschko:
yeah
let me check it today or tomorrow

in theory its only a porlet feature, not really required by JSF TCK

tandraschko:
@werpu shouldnt this be enough? https://github.com/apache/myfaces/commit/12be5de374101658f733ce87e28a378990a6730b

please give it a try

tandraschko:
containsKey uses getAttribute, see AbstractAttributeMap

werpu:
I will run the tests, but the HTMLResponseStateManager definitiely still needs some of the fixes I mentioned, because it is responsible for rendering the prefixed names!
The ugly thing is, while we have an api which determines the NamingContainer name, there are some phase where it is not present yet (hence the hack with the parameter lookup) and to the worse it might change later in a navigational case (if possible)

tandraschko:
i will commit the writing part now
if the TCK is green, i'm fine with the current changes
everything else should be in the porlet adapter code then

werpu:
Yes if the TCk passes then we are set

Thomas Andraschko ***@***.***> schrieb am Do., 2. Feb. 2023,
17:01:

werpu:
closing this one, we will take the alternative solution stemming partially from this one partially from @tandraschko s code!

PR #504, opened 2023-01-27
MYFACES-4553: Fix WELD-001303 By Forcing Flow Transition Earlier 

volosied:
Fix for https://issues.apache.org/jira/browse/MYFACES-4553

Details can be found in the comment in the JIRA above.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/application/NavigationHandlerImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/application/NavigationHandlerImpl.java
@@ -101,6 +101,8 @@ public class NavigationHandlerImpl extends ConfigurableNavigationHandler
     
     private ViewIdSupport viewIdSupport;
 
+    public final String STARTED_FLOW_TRANSITION = "STARTED_FLOW_TRANSITION";
+
     public NavigationHandlerImpl()
     {
         if (log.isLoggable(Level.FINEST))
@@ -364,6 +366,11 @@ public class NavigationHandlerImpl extends ConfigurableNavigationHandler
             }
         }
     }
+
+    private Boolean didFlowTransitionAlready(FacesContext facesContext)
+    {
+        return facesContext.getAttributes().containsKey(STARTED_FLOW_TRANSITION);
+    }
     
     private void applyFlowTransition(FacesContext facesContext, NavigationContext navigationContext)
     {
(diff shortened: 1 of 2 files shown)


volosied:
This code addresses the issue found in the TCK, but it fails in our tests. You can find the failures in the build log, for example: 
```
Error:  Tests run: 5, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 2.055 s <<< FAILURE! - in org.apache.myfaces.application.flow.FlowMyFacesCDIRequestTestCase
Error:  org.apache.myfaces.application.flow.FlowMyFacesCDIRequestTestCase.testFlow1_1  Time elapsed: 0.369 s  <<< FAILURE!
java.lang.AssertionError: 
expected null, but was:<<navigation-case>
  <from-view-id>/flow1_1.xhtml</from-view-id>
  <from-outcome>content</from-outcome>
  <to-view-id>/flow1/content.xhtml</to-view-id>
</navigation-case>>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotNull(Assert.java:756)
	at org.junit.Assert.assertNull(Assert.java:738)
	at org.junit.Assert.assertNull(Assert.java:748)
	at org.apache.myfaces.application.flow.FlowMyFacesCDIRequestTestCase.testFlow1_1(FlowMyFacesCDIRequestTestCase.java:72)
```

https://github.com/apache/myfaces/blob/93a47f3c78716f0ff3c01227ee1cd5972aeb965b/impl/src/test/java/org/apache/myfaces/application/flow/FlowMyFacesCDIRequestTestCase.java#L59-L72

volosied:
```
Error:  Failures: 
Error:    FlowMyFacesRequestTestCase.testFlow1_10:940 expected:<flow[3]> but was:<flow[1]>
Error:    FlowMyFacesRequestTestCase.testFlow1_10_1:1028 expected:<flow[3]> but was:<flow[1]>
Error:    FlowMyFacesRequestTestCase.testFlow1_10_2:1116 expected:<flow[3]> but was:<flow[1]>
Error:    FlowMyFacesRequestTestCase.testFlow1_8:684 expected:<flow[3]> but was:<flow[1]>
```

Made a update and it's down to 4 failures now. However, I don't really like this approach, but I don't see much we can do. We follow the spec (mostly), but I think Mojarra just made more updates made to address edge cases and the spec was never updated.

Basically, when should a flow be entered so that flowScope is active?

melloware:
I wonder if this is a question for BalusC?

volosied:
GREEN GREEN GREEN!  💚   Never been so happy to see a "BUILD SUCCESSFUL". 

Still need do some more testing and refactor the changes a bit, but I think I've fixed this without causing any regressions.

melloware:
Congratulations!!  I know you have been fighting this one!

impl/src/main/java/org/apache/myfaces/application/NavigationHandlerImpl.java
```diff
@@ -735,17 +709,59 @@ public NavigationCase getNavigationCommand(
...
+                                    // Ask the FlowHandler for a Flow for this flowId, flowDocumentId pair. Obtain a 
+                                    // reference to the start node and execute this algorithm again, on that start node.
+                                    complete = true;
+                                }
+                            }
+                            if (!complete && flowNode instanceof SwitchNode)
+                            {
+                                outcomeToGo = calculateSwitchOutcome(facesContext, (SwitchNode) flowNode);
+                                // Start over again checking if the node exists.
+                                //fromAction = currentFlow.getId();
```
tandraschko, line 741: commented code? remove?!

melloware (approved):
Looks good but I agree with Thomas's feedback

impl/src/main/java/org/apache/myfaces/application/NavigationHandlerImpl.java
volosied: ah, it was originally there, but let me remove it.

PR #505, opened 2023-01-28
MYFACES-4554: log FINEST rather than INFO

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4554


melloware (approved):

tandraschko:
Please see my jira comment, doesnt make sense IMO

PR #508, opened 2023-01-31
FacesConfig#version() cannot return null

KyleAure:
Related to https://github.com/jakartaee/faces/issues/1784
Mojarra already fixed: https://github.com/eclipse-ee4j/mojarra/pull/5210

We need to return a valid value and not null. 
Mojarra also backported this to their 4.0 release. MyFaces might also want to do the same.


melloware:
I think @tandraschko already made this fix in 4.0?

KyleAure:
Oops looks like it was already fixed yesterday :) Closing

PR #510, opened 2023-02-02
MYFACES-4533: JS only for a better solution

werpu:
By suggestion of @tandraschko he wants to work on a better solution for the issue!

Providing a pull request against main for the js only changes, so that he can take over with a hopefully better solution to the issue!

PR #511, opened 2023-02-06
JavaScript: Check if getAttribute is a function before calling it (MYFACES-4533 follow up)

volosied:
With the MYFACES-4533 fix (#510) merged in, we tries to run our test suite against it and found this htmlunit/javascipt error hitting numerous tests:

`Caused by: net.sourceforge.htmlunit.corejs.javascript.EcmaError: TypeError: Cannot find function getAttribute in object [object Text]. (http://localhost:8010/JSF22Miscellaneous/jakarta.faces.resource/faces.js.xhtml?ln=jakarta.faces&stage=Development#5213) `

Manual tests via web browser show no errors so this problem looks to be just another HTMLUnit bug. 

As a workaround for now, I'd like to test if this is a function so that these tests can pass.

### DIFF
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxUtils.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxUtils.js
@@ -186,7 +186,10 @@ _MF_SINGLTN(_PFX_XHR+"_AjaxUtils", _MF_OBJECT,
         form = this._Dom.byId(form);
         var _t = this;
         var foundNames = this._Dom.findAll(form, function(node) {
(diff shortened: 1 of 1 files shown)


volosied:
@werpu  Any objection to this change? It should not affect anything, but would help us out by avoid the error in the description (also might assist anyone else using HTMLUnit).

werpu:
Hello absolutely no objection.

Go ahead

Sorry, that it took me so long to answer, but I have been busy in another
project.

Werner


Am Di., 7. Feb. 2023 um 18:13 Uhr schrieb Volodymyr Siedlecki <
***@***.***>:

volosied:
Thanks! :)

PR #513, opened 2023-02-09
MYFACES-4551 follow up to fix radio selection

volosied:
The original fix for MYFACES-4551 ( #492) created some regressions, so I'm adding a follow up patch here.

The issue was with where the value expression was placed. 

For example, if multiple selectOneRadio elements are used and the the VE is in first element, and spec says to use the first one in the group with a VE ([see here -- group](https://jakarta.ee/specifications/faces/4.0/renderkitdoc/HTML_BASIC/jakarta.faces.SelectOnejakarta.faces.Radio.html)). If someone picked the second radio button (Value2),  [our code sets the submitted value on the first element with the VE.](https://github.com/apache/myfaces/blob/6bf7e9d86d881165d79f39927d3abd2d80bc76ff/impl/src/main/java/org/apache/myfaces/renderkit/html/util/HtmlRendererUtils.java#L372-L379) However, with my last fix, during validation of the first selectOneRadio element doesn't have Value2 (only Value 1 is contained) so it fails validation.  This is because of the use of this within`SelectItemsIterator(this, context)` in the uiselectone#validateValue method. 

This fix has the group aggregated for the submitted value to be searched ( There was also another problem in our original code (before  MYFACES-4551) since didn't convert the submitted value. Now we do. )

```
        <h:form id="form1" prependId="false">
            <h:selectOneRadio id="radio0" group="testGroup" value="#{selectOneRadioGroupBean.selectedValue}">
                <f:selectItem itemValue="Value1"/>
            </h:selectOneRadio>
            <h:selectOneRadio id="radio1" group="testGroup">
                <f:selectItem itemValue="Value2"/>
            </h:selectOneRadio>
            <h:selectOneRadio id="radio2" group="testGroup">
                <f:selectItem itemValue="Value3"/>
            </h:selectOneRadio>
            <h:commandButton id="submitButton" value="Submit"/>
        </h:form>
 ```

### DIFF
--- a/api/src/main/java/jakarta/faces/component/UISelectOne.java
+++ b/api/src/main/java/jakarta/faces/component/UISelectOne.java
@@ -81,7 +81,7 @@ public class UISelectOne extends UIInput
         String group = getGroup();
         String submittedValue = (String) getSubmittedValue();
 
-        if (group != null && !group.isEmpty()) 
+        if (group != null && !group.isEmpty() && isEmpty(submittedValue) ) 
         {
             final UIComponent form = getRadioNestingForm(context, this);
 
@@ -95,7 +95,7 @@ public class UISelectOne extends UIInput
                     {
                         // check if the is empty (see ) or if it's not valid (means this path has been taken already)
                         // See conditions listed under spec: uiselectone#processValidators
-                        if(isEmpty(submittedValue) && isSubmittedAlready(target)){
+                        if(isSubmittedAlready(target)){
                             previouslySubmittedOrValidated = true;
                             return VisitResult.COMPLETE;
                         }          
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by volosied** [comment]
Hmm. Need to investigate: 
```
Error:    UISelectOneTest.testValidateNotRequiredNotSelectOption:179 expected: <true> but was: <false>
Error:    UISelectOneTest.testValidateNotRequiredValid:87 expected: <true> but was: <false>
```

File: api/src/main/java/jakarta/faces/component/UISelectOne.java [hunk 1]
```diff
@@ -95,23 +95,24 @@ public VisitResult visit(VisitContext visitContext, UIComponent target)
...
-            // Skip further validation due to either 
-            // 1) one of the submissions are not valid (for instance, required, but none submitted (see jakarta/faces#329))
-            // 2) submitted value has been found and validated
-            return;
+            if(previouslySubmittedOrValidated){
+                // Skip further validation due to either 
+                // 1) one of the submissions are not valid (for instance, required, but none submitted (see jakarta/faces#329))
+                // 2) submitted value has been found and validated
+                return;
+            }
```
volosied, line 112: Did some refactoring here as we should really only return when `isEmpty(submittedValue)` is true since we're just trying to skip further validation.

melloware (approved):

PR #514, opened 2023-02-10
Feature/myfaces 4466: Merge Pull request into main (RC5 Target) - Faces.js new scripts!

werpu:
I have to open another pull request on the same branch, because the original one has been showing a merge error for quite some time. The merge is imminent, I just want to run another set of TCKs tests with the latest codebase


werpu:
For the sake of documentation, I will close the old pull request, because it has been broken for quite a while.
So that nothing gets lost documentationwise, here is the link:
https://github.com/apache/myfaces/pull/356

(The old one will be closed now)

werpu:
need to check why it is failing the ajax integration test!

werpu:
TCKs pass, however the Tobago people reported an issue with the fileupload for the RC34 of the jsf_ts codebase which needs a fix, so no merge this week, RC33 worked, but I do not want to fork the codebase there.

werpu:
Integration Tests pass:
image

TCK:
Faces23 / Ajax:
image

Faces 22
image

Faces 23/ajax: 
image


Namespace:
image

Note originally in faces 23 Test 1423 was failing, this is an issue of the test itself (timing issue, with a race condition between the webdriver and the test code, which is not fixable on framework level, but the test itself needs adaption)
The adapted test passes. I will issue a fix for the test on the TCK side on monday together with a small selenium test framework update!

werpu:
Hi bad news... i got a mild infection, and currently being in bed with fever, so I will leave the honor for the merge to @volosied, @tandraschko or @pnicolucci unless you want to wait until I have recovered (which might be wednesday)
Feel free to go ahead with the merge!

werpu:
Hi, given, I am still somewhat weak from the fever at the weekend, I will keep it short: 

This is mostly a complexity reduction, because instead of 4-5 different target builds (without mapping files back then) we had in the original, we now have only two: production and development, but also we have mapping files. Thats all there is to it.

tandraschko:
lets merge this
thanks for your great and hard work werner!

melloware:
I will second that I know this was a LOT of work.

tandraschko:
the build somehow fails on localhost:

> myfaces.faces.ajax@1.0.0 test
> env TS_NODE_PROJECT=./tsconfig-myfaces.json mocha

Der Befehl "env" ist entweder falsch geschrieben oder konnte nicht gefunden werden.

any idea?

werpu:
@tandraschko could you resolve the issue? Otherwise I will look into it!

tandraschko:
@werpu i didnt had the time 
if you can "fix" our code, i can just do a mvn install again on my machine to test it

werpu:
resolved it with @tandraschko, we can see the issue as closed now. The fix was indeed to use the cross-env package as replacement for the env command. I was not aware of the cross platform issues env in Node has and many people aren´t as it seems.

PR #520, opened 2023-02-13
MYFACES-4561 Change Context Parameter Logging

volosied:
https://issues.apache.org/jira/browse/MYFACES-4561

Although the hope was to not log these context param messages in production. That change would have unintended consequences since the params then won't be logged in at all.  This can make it difficult to debug applications if there's no trace regarding the context parameters. 

I add new logging and refactored the code, so the new behavior is now: 

auto - logs in dev mode only
true - logs in production and dev mode
false - no logging in either mode (fine messages still logged when trace is specified -- added for support)

Other project stage modes are not considered. 

Additionally,  _found_ context parameters are now logged.

### DIFF
--- a/impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
+++ b/impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
@@ -54,56 +54,56 @@ public class WebConfigParamsLogger
 ## -------------------------------
 ##
 #macro (writePropertyCheck $webConfigParam)
-            paramValue = facesContext.getExternalContext().getInitParameter("$webConfigParam.name");
-            if (paramValue == null)
-            {
-                logMessageToAppropriateLevel("No context init parameter '$webConfigParam.name' found#if($webConfigParam.defaultValue), using default value '$webConfigParam.defaultValue'#end.");
-            }
+        paramValue = facesContext.getExternalContext().getInitParameter("$webConfigParam.name");
+        if (paramValue == null)
+        {
+            logMessageToAppropriateLevel("No context init parameter '$webConfigParam.name' found#if($webConfigParam.defaultValue), using default value '$webConfigParam.defaultValue'#end.");
+        }
 #if (!$webConfigParam.isDeprecated())
 #if ($webConfigParam.expectedValues)
-            else
+        else
+        {
+            boolean found = false;
+            String[] expectedValues = StringUtils.trim(StringUtils.splitShortString("$webConfigParam.expectedValues",','));
+            for (int i = 0; i < expectedValues.length; i++)
             {
-                boolean found = false;
-                String[] expectedValues = StringUtils.trim(StringUtils.splitShortString("$webConfigParam.expectedValues",','));
-                for (int i = 0; i < expectedValues.length; i++)
-                {
 #if ($webConfigParam.isIgnoreUpperLowerCase())
-                    if (paramValue.equalsIgnoreCase(expectedValues[i]))
+                if (paramValue.equalsIgnoreCase(expectedValues[i]))
 #else
-                    if (paramValue.equals(expectedValues[i]))
+                if (paramValue.equals(expectedValues[i]))
 #end
-                    {
-                        found = true;
-                        break;
-                    }
-                }
-                if (!found)
                 {
-                    if (log.isLoggable(Level.WARNING))
-                    {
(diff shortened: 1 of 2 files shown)


tandraschko:
why not just redefine 
LOG_WEB_CONTEXT_PARAMS_DEFAULT = "auto";

to "auto log in development stage"?

false = never
true = always

volosied:
I can keep the new logging code as is, but undo my LOG_WEB_CONTEXT_PARAMS changes.

Although, "auto" doesn't seem to make much sense since now it only applies to dev mode (true / false is enough IMO, default true), and I still would like to have fine logging simply for support.

volosied:
I think I'm following you now.

So here's the new behavior for these _info_ message: 
auto -  logs in dev mode only (not prod) (fine messages still logged in prod when trace is specified)
true - logs in production and dev mode  (fine not needed, due to info messages)
false - no logging in either  (fine messages still logged in prod when trace is specified)

impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
```diff
@@ -129,7 +129,18 @@ public class WebConfigParamsLogger
...
+    private static void logMessageToAppropriateLevel(String text)
+    {
+        if(LOG_WEB_CONTEXT_PARAM && log.isLoggable(Level.INFO))
+        {
+            log.info(text);
+        }
+        else if (log.isLoggable(Level.FINEST))
+        {
+            log.finest(text);
+        }
```
volosied, line 144: Helper function to avoid repetition

impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
```diff
@@ -34,88 +34,88 @@ public class WebConfigParamsLogger
...
-
-        if (log.isLoggable(Level.INFO))
+        if(LOG_WEB_CONTEXT_PARAM)
         {
             log.info("Scanning for context init parameters not defined. It is not necessary to define them all into your web.xml, " +
-                     "they are just provided here for informative purposes. To disable this messages set " +
-                     MyfacesConfig.LOG_WEB_CONTEXT_PARAMS + " config param to 'false'");
+                "they are just provided here for informative purposes. To disable this messages set " +
+                MyfacesConfig.LOG_WEB_CONTEXT_PARAMS + " config param to 'false'");
+        }
```
volosied, line 118: Not really needed, so it's only logged when LOG_WEB_CONTEXT_PARAM is true.

impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
```diff
@@ -34,88 +34,88 @@ public class WebConfigParamsLogger
...
+            }
+            catch(Exception e)
+            {
+                if (log.isLoggable(Level.WARNING))
                 {
-                    if (log.isLoggable(Level.WARNING))
-                    {
-                        log.warning("Wrong value in context init parameter '$webConfigParam.name' (='" + paramValue + "'), using default value #if($webConfigParam.defaultValue)'$webConfigParam.defaultValue'#else'null'#end");
-                    }
+                    log.warning("Wrong value in context init parameter '$webConfigParam.name' (='" + paramValue + "'), using default value #if($webConfigParam.defaultValue)'$webConfigParam.defaultValue'#else'null'#end");
```
volosied, line 103: Majority of the "changes" are just me indenting this to the left once since I removed an if statement this code was nested in.

volosied:
@tandraschko Let me know what you think of the new changes.

impl/src/main/resources/META-INF/WebConfigParamsLogger.vm
```diff
@@ -34,88 +34,88 @@ public class WebConfigParamsLogger
...
-        if (!myfacesConfig.isLogWebContextParams())
-        {
-            //No log if is disabled or is in auto mode and project stage is UnitTest or SystemTest
-            return;
-        }
+        LOG_WEB_CONTEXT_PARAM = myfacesConfig.isLogWebContextParams();
 
-
-        if(myfacesConfig.isRiImplAvailable() && myfacesConfig.isMyfacesImplAvailable())
+        if(myfacesConfig.isRiImplAvailable() && myfacesConfig.isMyfacesImplAvailable() && log.isLoggable(Level.SEVERE))
```
tandraschko, line 45: severe should never be disabled TBH

volosied: I've removed this change.

bohmber:
@volosied I think auto is a bad name. Why not dev-only or reduced or something more descriptive

PR #525, opened 2023-02-16
Update LOG_WEB_CONTEXT_PARAMS options

volosied:
https://github.com/apache/myfaces/pull/520#issuecomment-1432835186

Updating to "always, dev-only, never" so that they describe the functionality better.  (Note: always only means dev or production and not the other stages)

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
+++ b/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
@@ -714,14 +714,15 @@ public class MyfacesConfig
     /**
      * Indicate if INFO logging of all the web config params should be done before initialize the webapp. 
      * <p>
-     * If is set in "auto" mode, web config params are only logged in "Development" mode. 
-     * If is set to "true", web config params are only logged in "Production" and "Development" mode. 
-     * If is set in "false" mode, no info logging occurs in either mode.
+     * If set to "dev-only" mode, web config params are only logged in "Development" mode. 
+     * If set to "true" or "always", web config params are only logged in "Production" and "Development" mode. 
+     * If set to "false" or "never", no info logging occurs in either mode.
+     * If set to "non-production", logging occurs in all project stages except production.
      * </p> 
      */
-    @JSFWebConfigParam(expectedValues="true, auto, false", defaultValue="auto")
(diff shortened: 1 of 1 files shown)


volosied:
We don't log anything for UnitTest and SystemTest. That should stay the same, right?

bohmber:
I prefer keep true, false as well for backwards compatiblity. That means always, dev-only, never, true, false

bohmber:
UnitTest, SystemTest good question. What about none-production or something similar

volosied:
@bohmber I've updated the code again. Let me know what you think. 

The 5 options now are: always, true, dev-only, never, false, non-production

bohmber (approved):
For me it looks better now. Lets wait for @tandraschko comments

volosied:
Spotted a new errors -- they're fixed now.

bohmber (approved):

impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
```diff
@@ -714,14 +714,15 @@
     /**
      * Indicate if INFO logging of all the web config params should be done before initialize the webapp. 
      * <p>
-     * If is set in "auto" mode, web config params are only logged in "Development" mode. 
-     * If is set to "true", web config params are only logged in "Production" and "Development" mode. 
-     * If is set in "false" mode, no info logging occurs in either mode.
+     * If set to "dev-only" mode, web config params are only logged in "Development" mode. 
+     * If set to "true" or "always", web config params are only logged in "Production" and "Development" mode. 
+     * If set to "false" or "never", no info logging occurs in either mode.
```
volosied, line 719: Maybe? 
```suggestion
     * If set to "false" or "never",  info logging does not occur in any mode.
```

bohmber: That is up to you

volosied:
No word from Thomas yet so I set it to dev-only here: https://github.com/apache/myfaces/commit/381f2714da9e32ea23a77df20db5983afcc7408d

Will start RC5 release now.

tandraschko:
TBH it feels a bit overengineered now with to many options for such a small logging

But If its only logged in dev stage per default, im ok with it
Please dont forget the homepage

volosied:
I agree with Thomas. I've already changed auto to dev-only, so I'll just close this PR.

PR #531, opened 2023-02-21
MYFACES-4570: fix NPE

pnicolucci:
https://issues.apache.org/jira/browse/MYFACES-4570

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/resource/FacesJSResourceLoader.java
+++ b/impl/src/main/java/org/apache/myfaces/resource/FacesJSResourceLoader.java
@@ -95,7 +95,7 @@ public class FacesJSResourceLoader extends ResourceLoaderWrapper
                 getHttpServletRequest(FacesContext.getCurrentInstance().getExternalContext());
 
         //name mapping does not happen on meta level
-        if(!(libraryName.equals(JsfLibrary.NAMESPACE) &&
+        if(!(JsfLibrary.NAMESPACE.equals(libraryName) &&
                 (
                  resourceName.contains(facesJsShortName) ||
                  resourceName.contains(facesJsLongName) //  just in case the behavior is altered in the future


tandraschko:
+1
we could do a Objects.equals instead?

pnicolucci:
> +1 we could do a Objects.equals instead?

sure, let me push up a change.

pnicolucci:
fyi looks like our integration tests are broken (unrelated to this PR):

`Error:    IntegrationTest.testPlainHtml » Runtime Unable to instantiate Drone via org.openqa.selenium.chrome.ChromeDriver(ChromeOptions): org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: session not created: This version of ChromeDriver only supports Chrome version 108`

pnicolucci:
@tandraschko  look better? Thanks for the review!

impl/src/main/java/org/apache/myfaces/resource/FacesJSResourceLoader.java
```diff
@@ -95,7 +96,7 @@ public InputStream getResourceInputStream(ResourceMeta resourceMeta)
                 getHttpServletRequest(FacesContext.getCurrentInstance().getExternalContext());
 
         //name mapping does not happen on meta level
-        if(!(libraryName.equals(JsfLibrary.NAMESPACE) &&
+        if(!Objects.equals(libraryName,null) && !(libraryName.equals(JsfLibrary.NAMESPACE) &&
```
melloware, line 99:
```suggestion
        if(!JsfLibrary.NAMESPACE.equals(libraryName)) &&
```

Joshua Bloch always recommends flipping to use the Constant first this way you can't have an NPE

melloware (changes requested):

pnicolucci:
@melloware should have been obvious to me :) Thanks! Look good now?

melloware (approved):

melloware:
Yep and I sent @bohmber a note its saying something about Drone only supporting Chrome 108 and I bet GitHub Actions Linux is is on Chrome 110.  I don't fully understand the Drone stuff.

PR #536, opened 2023-02-26
MYFACES-4573: 4.0 Quarkus native exception registration

melloware:
1. register web.xml
1. register all Exceptions found in web.xml for reflection (always include ViewExpiredException no matter what)

Since there is no ExternalContext but i needed to use WebXmlParser I made as little changes as possible to allow it to be used in the Deployment processor.


tandraschko:
looks ok for me

PR #538, opened 2023-02-28
MYFACES-4563: ignore viewMap in debug to avoid OOM

volosied:
https://issues.apache.org/jira/browse/MYFACES-4563?


tandraschko:
+1

PR #544, opened 2023-03-05
MYFACES-4581: Felix 5.1.8

melloware:
Fixes ConcurrentModificationException trying to build locally.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -195,13 +195,10 @@
                     <version>1.9.1</version>
                 </plugin>
 
-                <!-- Version 4.1.0 and higher does not output the version ranges for some packages
-                     when specified by <Import-Package> in API/IMPL plugins
-                     See https://issues.apache.org/jira/browse/FELIX-6566 -->
                 <plugin>
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
-                    <version>4.0.0</version>
+                    <version>5.1.8</version>
                 </plugin>
 
                 <plugin>


melloware:
@volosied I didn't understand https://issues.apache.org/jira/browse/FELIX-6566 which is still open can you see if version 5.1.8 still has the issue you refer to?  I cannot build 4.0 locally without this plugin upgrade

volosied:
Building with 5.18 produces the following impl manifest: 

```
Import-Package: jakarta.annotation;version="[2.1.0,3.0.0)",
javax.crypto
javax.crypto.spec
jakarta.ejb;version="[4.0.0,5.0.0)";resolution:=optional
jakarta.el
javax.naming
jakarta.persistence;version="[3.1.0,4.0.0)";resolution:=optional
jakarta.servlet
jakarta.servlet.http
jakarta.servlet.annotation;resolution:=optional
javax.xml.parsers
```

But we specify version here: 
https://github.com/apache/myfaces/blob/main/impl/pom.xml#L632-L646


Looks like it still occurs. Let me look at this again.

melloware:
OK thanks for looking at this.

volosied:
Even updating to the jakarta spec jars doesn't seem to fix the jakarta.el import: 

https://github.com/volosied/myfaces/pull/new/jakarta-dep-update

volosied:
I'll try to reach out to some more people.

melloware:
We use Felix 5.1.8 in PrimeFaces too and I don't think we have had issues so this is strange.

tandraschko:
any news @volosied?

tandraschko:
i need to merge this for now as all CI are failing.

PR #545, opened 2023-03-06
MYFACES-4582: Quarkus hot reload

melloware:
@tandraschko this works but still generates this warning:

```
Called FactoryFinderProviderFactory.setFactory after initialized FactoryFinder (first call to getFactory() or setFactory()). This method should be called before any 'web context' is initialized in the current 'classloader context'. By that reason it will not be changed.
```

But everything seems to be working and hot reload works again.

I had to remove the unmodifiable map or else it also throws an exception trying to add the models.

### DIFF
--- a/extensions/quarkus/runtime/src/main/java/org/apache/myfaces/core/extensions/quarkus/runtime/QuarkusFacesInitializer.java
+++ b/extensions/quarkus/runtime/src/main/java/org/apache/myfaces/core/extensions/quarkus/runtime/QuarkusFacesInitializer.java
@@ -36,28 +36,30 @@ import org.apache.myfaces.core.extensions.quarkus.runtime.spi.QuarkusFactoryFind
  */
 public class QuarkusFacesInitializer extends FacesInitializerImpl
 {
+    // DIRTY HACK as somehow quarkus initializes a second time?
     private static boolean initialized = false;
-    
+
     @Override
     public void initFaces(ServletContext servletContext)
     {
-        // DIRTY HACK as somehow quarkus initializes a second time?
         if (!initialized)
         {
-            initialized = true;
-
+            // only initialize FactoryFinderProviderFactory once and not on hot reloading
             FactoryFinderProviderFactory.setInstance(new QuarkusFactoryFinderProviderFactory());
-
-            // see FacesDataModelExtension
-            FacesDataModelManager facesDataModelManager = CDI.current().select(FacesDataModelManager.class).get();
-            for (Map.Entry<Class<? extends DataModel>, Class<?>> typeInfo
-                    : MyFacesRecorder.FACES_DATA_MODELS.entrySet())
-            {
-                facesDataModelManager.addFacesDataModel(typeInfo.getValue(), typeInfo.getKey());
-            }
-            facesDataModelManager.init();
-
-            super.initFaces(servletContext);
+            initialized = true;
+            return;
         }
+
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/apache/myfaces/cdi/model/FacesDataModelManager.java
```diff
@@ -58,10 +58,6 @@ public void addFacesDataModel(Class<?> forClass, Class<? extends DataModel> data
 
     public void init()
```
tandraschko, line 59: we should add a reinit method here instead of modifiy the existing logic

extensions/quarkus/runtime/src/main/java/org/apache/myfaces/core/extensions/quarkus/runtime/QuarkusFacesInitializer.java
```diff
@@ -36,28 +36,21 @@
  */
 public class QuarkusFacesInitializer extends FacesInitializerImpl
 {
-    private static boolean initialized = false;
     
     @Override
     public void initFaces(ServletContext servletContext)
     {
-        // DIRTY HACK as somehow quarkus initializes a second time?
```
tandraschko, line 44: doesnt myfaces start 2 times with this change now? or did they fix it?

melloware: I only see it start once with that warning about FactoryFinder?

impl/src/main/java/org/apache/myfaces/cdi/model/FacesDataModelManager.java
melloware: good call.

extensions/quarkus/runtime/src/main/java/org/apache/myfaces/core/extensions/quarkus/runtime/QuarkusFacesInitializer.java
tandraschko: is public void initFaces(ServletContext servletContext) called 2 times?

melloware: I am checking.

melloware: OK it does initialize twice and again every time hot reload happens so my latest change make it work on initial startup and on hot reload with no more warnings!

impl/src/main/java/org/apache/myfaces/cdi/model/FacesDataModelManager.java
```diff
@@ -64,6 +64,11 @@ public void init()
         }
     }
 
+    public void reinitialize()
```
tandraschko, line 67: maybe better call it reset and just do: facesDataModels = null;

tandraschko:
see my last comment; then +1

impl/src/main/java/org/apache/myfaces/cdi/model/FacesDataModelManager.java
melloware: Done

PR #546, opened 2023-03-08
MYFACES-4588

werpu:
Fixes the issue that getWindowId did not look into the document in the correct order.
Also added tests for the case and added tests for the file upload given from the Tobago guys (which I forgot to add to git)


werpu:
Seems like we have an integration test build issue, apparently the chrome driver has a version mismatch with chrome version in the build system. Wont merge until everything passes. But the fix is fine (see the test protocols) and independent of the failing integration tests.

bohmber:
Unfortunately there is not better fix at the moment

bohmber:
The version should be 110.0.5481.77

api/src/client/typescript/faces/impl/AjaxImpl.ts
```diff
@@ -54,7 +54,7 @@ import {
     EMPTY_STR,
     CTX_PARAM_MF_INTERNAL,
     NAMED_VIEWROOT,
-    NAMING_CONTAINER_ID
+    NAMING_CONTAINER_ID, $nsp
```
volosied, line 57: Should it be on a new line?

volosied:
+1  

Tested locally and the fix works! Thanks :)

api/src/client/typescript/faces/impl/AjaxImpl.ts
werpu: fixed

werpu:
ci build still failing, but given this is unrelated to my fix and also on the master branch, I guess we can merge now!

bohmber:
Build is running successful now with the correct chromeDriverVersion

werpu:
Thanks for fixing it!

PR #547, opened 2023-03-09
https://issues.apache.org/jira/browse/MYFACES-4584: Improvements in the Ajax Codebase

werpu:
- Streams are now entirely eliminated in favor of JS Arrays (very likely faster because native arrays are handled by the browser)
- Added a non intrusive Array shim for flatMap (some browsers in our baseline do not support it)
- Reduced internal Config data structure to raw associative arrays gone are now the configs and subconfigs (easier debugging probably faster also)
- Simplified the XHRQueue code it is now 50% smaller and less complex
- Cleaned up the XhrRequest class
- Moved the AsyncRunnable into an abstract base class 
- Improved error reporting in the unit tests, the error logs of wanted errors are now swallowed
- Added Es2019Array (non intrusive array shim) testing to our unit tests, parts of the unit tests now run in the shim, to expose if a non shimmed function is called!
- Several tests added, including an onLoad test, which was not triggered before due to a wrong filename!


melloware:
Nice work @werpu !

werpu:
I will leave it open for review until tomorrow, will merge sometime tomorrow!

melloware (approved):

werpu:
Merged!

Another "improvement" still possible is the elimination of configs, but given the latest changes in the internal config data stuctures, there is little to gain from it no code reduction, no complexity reduction and no speed gain, only a tad closer to the language itself, atm. 

So not worth the effort for now! In reality the configs just would be replaced by direct assoc arrays and function calls on top of them but the code basically stays the same mostly. 

So I guess this would not really reduce complexity (config.assign(key1,key2,key3).value = "foobaz" would become assign(assocArray, key1,key2,key3).value = "foobaz").

For debugging purposes - configs now use a pure associative arrays with data only anyway as internal data structures, shifting a little bit away from the pure recursive functional approach used before)
Either way should this change be wanted in the long run, the foundations for this are there (and used internally by the Config)

A shift to pure assoc arrays probably will make sense as soon as typescript allows constructs like
assocArray ?.key1?.key2?.key3 = "value"

atm this is a read only operation which forced me to introduce Configs in the code first place!

PR #548, opened 2023-03-09
Use Java11+ HTTP Client for Selenium

pnicolucci:
https://www.selenium.dev/blog/2022/using-java11-httpclient/#integrating-the-java-11-client

Resolves issues we're seeing with Chrome 111:
[ERROR]   IntegrationTest.test ▒ Runtime Unable to instantiate Drone via org.openqa.selenium.chrome.ChromeDriver(ChromeOptions): org.openqa.selenium.remote.http.ConnectionFailedException: Unable to establish websocket connection

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -68,6 +68,9 @@
                             <includes>
                                 <include>**/*IntegrationTest</include>
                             </includes>
+                            <systemPropertyVariables>
+                                <webdriver.http.factory>jdk-http-client</webdriver.http.factory>
+                            </systemPropertyVariables>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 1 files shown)


pnicolucci:
@werpu , I'm hitting the same issue you reported/fixed in `https://github.com/jakartaee/faces/pull/1799` with the MyFaces integration tests locally. I couldn't see where we have a similar code to how you fixed this in the TCK. However, these changes got me up and running. What do you think? Is there a way to make the same change that was made in the TCK or should we go with what I've proposed here?

bohmber:
@pnicolucci thanks for the suggestion will be used in #549

werpu:
Hi sorry to jump in here so late.

Hi following fixed it for me  options.addArguments("--remote-allow-origins=*");
aka chrome options remote-allow-origins=*

This basically reduces the security of chrome on the latest builds to allow websocket connections again, which seems to be the issue why the tests fail with the remote connection error on the chrome driver!
Not sure where to add this in our myfaces tests, this is probably a selenium setting you can pass down to the webdriver!
(On the jakarta codebase I instantiate the selenium webdriver programmatically hence I put it into the code itself)

PR #551, opened 2023-03-10
https://issues.apache.org/jira/browse/MYFACES-4593

werpu:
Fix and new tests for https://issues.apache.org/jira/browse/MYFACES-4593: Fixes a small deviance which was not triggered from the spec jsdocs!


werpu:
TCK (limited tests, only the ajax parts) Pass
personal integrationt tests pass
Unit Tests pass!

PR #552, opened 2023-03-14
MYFACES-4561

volosied:
https://issues.apache.org/jira/browse/MYFACES-4594


tandraschko:
looks good for me, just one task fails
can you check and merge?

PR #557, opened 2023-03-27
MYFACES-4596: Style tags must retain position and attributes during eval!

werpu:
Fix coming in from the Tobago/mona-dish downstream Update the latest dependencies!


werpu:
Tests pass, additional unit test for the tobago usecase has been added!

PR #567, opened 2023-04-09
MYFACES-4598: Quarkus put "jakarta.faces" proxies back in

melloware:
cc @tandraschko you were right it then was having reflection issues with Converters.


tandraschko:
Even other excludes will propably break EL

Try to produce a ArrayList in requestscope and access via EL

PR #568, opened 2023-04-09
MYFACES-4598: Revert  Quarkus reduce WARNING noise

melloware:


melloware:
@tandraschko you are right.  I guess there is nothing that can be done to reduce this noise on native image build.

PR #569, opened 2023-04-12
Quarkus 3.0.0 Final

melloware:
Quarkus 3.0 went final so I figured this is what we should compile against.


melloware:
failures are unrelatde to this change its more Drone -> Chrome issues.

PR #573, opened 2023-05-17
Apply MYFACES-4469 to 3.0.x

volosied:
https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4469

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/FacesConfigurator.java
+++ b/impl/src/main/java/org/apache/myfaces/config/FacesConfigurator.java
@@ -20,8 +20,6 @@ package org.apache.myfaces.config;
 
 import java.io.IOException;
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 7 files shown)


impl/src/main/java/org/apache/myfaces/config/FacesConfigurator.java
```diff
@@ -503,82 +480,57 @@ public void update()
...
-                return purgeMethod;
+                ((Purgeable) instance).purge();
             }
+
             if (instance instanceof FacesWrapper)
             {
-                instance = ((FacesWrapper)instance).getWrapped();
+                instance = ((FacesWrapper<T>) instance).getWrapped();
             }
         }
```
volosied, line 533: Returned originally missed, but added here: https://github.com/apache/myfaces/commit/218dd53ce0d38263fb74f6d49b70b5b15671eeeb

PR #590, opened 2023-06-23
MYFACES-4609: Web.xml init params evaluate for EL expressions

melloware:
This was my idea was to refactor this where the init params are being read to evaluate them for EL.

However I suspect using FacesContext here is not correct?

```java
        FacesContext facesContext = FacesContext.getCurrentInstance();
        return (T) facesContext.getApplication().evaluateExpressionGet(facesContext, param, Object.class);
```


tandraschko:
why is it not correct use FC? not available yet?

tandraschko:
i general i think its a nice feature but MyfacesConfig is currently handled AppScoped, to correctly support it, we would need to be request scoped (EL can change each request)
i would close it as wont fix for now TBH

melloware:
OK closing as wont fix!

PR #607, opened 2023-08-17
MYFACES-4623 check for duplicate ids

volosied:
Draft test fix for https://issues.apache.org/jira/browse/MYFACES-4623

Feel like a band-aid solution on top of another band-aid :/ 

This could use some refactoring to avoid code duplication?

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/view/facelets/PartialStateManagementStrategy.java
+++ b/impl/src/main/java/org/apache/myfaces/view/facelets/PartialStateManagementStrategy.java
@@ -52,6 +52,7 @@ import jakarta.faces.view.ViewMetadata;
 import org.apache.myfaces.application.StateManagerImpl;
 import org.apache.myfaces.application.viewstate.StateCacheUtils;
 import org.apache.myfaces.context.RequestViewContext;
+import org.apache.myfaces.core.api.shared.ComponentUtils;
 import org.apache.myfaces.util.lang.ClassUtils;
 import org.apache.myfaces.util.lang.HashMapUtils;
 import org.apache.myfaces.component.visit.MyFacesVisitHints;
@@ -642,17 +643,25 @@ public class PartialStateManagementStrategy extends StateManagementStrategy
                 {
                     try
                     {
-                        UIComponent parent = target;
-                        UIComponent dup = parent.getChildren().get(childIndex);
-                        if (child.getId().equals(dup.getId())) // avoid DuplicateIdException  - myfaces-4623
-                        {
-                            parent.getChildren().remove(childIndex.intValue());
-                            parent.getChildren().add(childIndex, child);
-                        }
-                        else
-                        {
-                             target.getChildren().add(childIndex, child);
-                        }
+                            UIComponent parent = target;
+                            for (int i = 0, childCount = parent.getChildCount(); i < childCount; i ++)
+                            {
+                                UIComponent dup = parent.getChildren().get(i);
+                                // myfaces-4623: avoid duplicate id exceptions w/ c:forEach for Resource Dependencies
+                                if (child.getId().startsWith(UIViewRoot.UNIQUE_ID_PREFIX + ComponentUtils.RD_ID_PREFIX) 
+                                        && child.getId().equals(dup.getId())) 
+                                {
+                                    // Replace
+                                    parent.getChildren().remove(i);
+                                    parent.getChildren().add(i, child);
+                                    done = true;
+                                    break;
+                                }
+                            }
+                            if(!done) // if not replaced yet, then just add it. 
+                            {
+                                target.getChildren().add(childIndex, child);
+                            }
                     }
                     catch (IndexOutOfBoundsException e)
                     {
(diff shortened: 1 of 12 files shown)


impl/src/main/java/org/apache/myfaces/view/facelets/PartialStateManagementStrategy.java
```diff
@@ -642,14 +642,26 @@ public void invokeContextCallback(FacesContext context, UIComponent target)
...
+                        else
+                        {
+                             target.getChildren().add(childIndex, child);
+                        }
                     }
                     catch (IndexOutOfBoundsException e)
                     {
                         // We can't be sure about where should be this 
                         // item, so just add it. 
+                         System.out.println("VS: ADDED");
```
melloware, line 661: Remove System.out.println but other than that LGTM

melloware (approved):

volosied:
Thanks - I'm gonna do some testing with this and if everything looks good, I'll merge next week. I'll also get a draft PR up soon for the other duplicate ID exception.

volosied:
I'll do TCK testing before any release, so I'll just merge this now now.

tandraschko:
i would like to have some tests before merging such a change

volosied:
Tried to create a simple scenario, but so far so luck.  This DupId Exception is also difficult to replicate easily. 

Requirements for this bug to occur are  

- Composite Component
   - c:forEach
     - p:calendar
     
 I'm not familiar with p:calendar but my suspicion is that the tree under it is creating elements which aren't marked. Thus during the restore phase, the children are simply added.  Specifically, it relates to the component resources created by p:calendar.  (target is ComponentResourceContainer and the child is jakarta.faces.component.UIOutput.  ) 
 
 This getChildList.add causes the duplicate id exception since restored child elements match an id of an existing element.  I believe the faulty element is `inputmask/inputmask.js.xhtml?ln=primefaces&v=13.0.0"`? 
 
 See error from the original test app: 
 
 ```
 org.apache.myfaces.view.facelets.compiler.DuplicateIdException: Component with duplicate id "j_id__rd_5" found. The first component is {Component-Path : [Class: jakarta.faces.component.UIViewRoot,ViewId: /test.xhtml][Class: org.apache.myfaces.component.ComponentResourceContainer,Id: jakarta_faces_location_head][Class: jakarta.faces.component.UIOutput,Id: j_id__rd_5]}
``` 
 
 
 I'll take another look later.

**Comment by volosied** [comment]
I'm not able to reproduce this outside the app. Perhaps you could take a look? I can't create another  custom component like p:calendar to replicate this problem with. @tandraschko 

 However.  c:forEach is broken and we've had many issues in the past.   I'm not too happy with this fix, but, given the state of things with c:forEach, I think we should merge this it.
 
 Leo also mentioned [here](https://github.com/apache/myfaces/blob/50d6e0ec3507f0b5d8c2b7492505ac683180b9a0/impl/src/main/java/org/apache/myfaces/view/facelets/PartialStateManagementStrategy.java#L604C1-L606C72): 
 
                     // By effect of c:forEach it is possible that the component can be duplicated
                    // in the component tree, so what we need to do as a fallback is replace the
                    // component in the spot with the restored version.
                    
                    
It seems to me that this very rare scenario with Resource Dependencies was overlooked?

**Comment by tandraschko** [comment]
i dont know what should be special on p:calendar? it just has `@ResourceDependency` like any other comp?

File: impl/src/main/java/org/apache/myfaces/view/facelets/PartialStateManagementStrategy.java [hunk 2]
```diff
@@ -642,14 +642,25 @@ public void invokeContextCallback(FacesContext context, UIComponent target)
                 {
                     try
                     {
-                        target.getChildren().add(childIndex, child);
+                        UIComponent parent = target;
+                        UIComponent dup = parent.getChildren().get(childIndex);
+                        if (child.getId().equals(dup.getId())) // avoid DuplicateIdException  - myfaces-4623
```
volosied, line 647: Maybe I should check for `dup.getAttributes().get(ComponentSupport.MARK_CREATED)`?

volosied: Actually, I cannot because of the initial check: `child.getAttributes().containsKey(ComponentSupport.MARK_CREATED)` is false.

melloware (approved):

volosied:
I've managed to reproduce it. I also realized my fix was not enough.  I can add a test, but I'm not sure how to do it via unit testing. Can I add it as an integration test?

volosied:
Test app: [myfaces-4623-test.zip](https://github.com/apache/myfaces/files/13044382/myfaces-4623-test.zip)

volosied:
@tandraschko Please look over and provide feedback. Also, let me know where I should add the test.

volosied:
@tandraschko  Just following up -- let me know your thoughts and the new code & where to place the test. Thanks.

tandraschko:
a integrationtests is ok for me, we just need some auto-testing to avoid breaking again in future

PR #615, opened 2023-10-03
Faces 5.0: Spec 1584 - Rely on CDI Class Scanning

volosied:
For `https://github.com/jakartaee/faces/issues/1584` 

Since CDI scanning is default now, I removed the `USE_CDI_FOR_ANNOTATION_SCANNING` configuration.

Question: Should we still have fallbacks 2 & 3?  

TODO: Create JIRA if one doesn't exist.  And investigate unit test failures...

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/annotation/DefaultAnnotationProvider.java
+++ b/impl/src/main/java/org/apache/myfaces/config/annotation/DefaultAnnotationProvider.java
@@ -128,21 +128,19 @@ public class DefaultAnnotationProvider extends AnnotationProvider
     @Override
     public Map<Class<? extends Annotation>, Set<Class<?>>> getAnnotatedClasses(ExternalContext ctx)
     {
-        if (MyfacesConfig.getCurrentInstance(ctx).isUseCdiForAnnotationScanning())
+        //1. Use CDI
+        BeanManager beanManager = CDIUtils.getBeanManager(ctx);
+        CdiAnnotationProviderExtension extension = CDIUtils.getOptional(beanManager,
+                CdiAnnotationProviderExtension.class);
+        if (extension != null)
         {
-            BeanManager beanManager = CDIUtils.getBeanManager(ctx);
-            CdiAnnotationProviderExtension extension = CDIUtils.getOptional(beanManager,
-                    CdiAnnotationProviderExtension.class);
-            if (extension != null)
-            {
-                return extension.getMap();
-            }
+            return extension.getMap();
         }
 
         Map<Class<? extends Annotation>, Set<Class<?>>> map = new HashMap<>();
         Collection<Class<?>> classes = null;
 
-        //1. Scan for annotations on /WEB-INF/classes
+        //2. Scan for annotations on /WEB-INF/classes
         try
         {
             classes = getAnnotatedWebInfClasses(ctx);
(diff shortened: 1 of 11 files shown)


volosied:
Down to 6 failures: 
```
[ERROR] Errors: 
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testBeanValidationNotAvailable:213 » NullPointer
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testDefaultBeanValidatorDisabled:186 » NullPointer
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testDefaultBeanValidatorDisabledButPresentInFacesConfig:153 » NullPointer
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testDefaultValidatorsClearedByLatterConfigFileWithEmptyElement:251 » NullPointer
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testDefaultValidatorsNotClearedByLatterConfigFileWithNoElement:292 » NullPointer
[ERROR]   FacesConfiguratorDefaultValidatorsTestCase.testDefaultValidatorsOverwrittenByLatterConfigFile:333 » NullPointer
```

I'm not sure how to enable CDI in the `AbstractJsfConfigurableMockTestCase` super class.

volosied:
Alternatively, we keep the config and switch it to false for the failng `FacesConfiguratorDefaultValidatorsTestCase` tests.

tandraschko:
if the spec switches to to this behavior, we should completely remove the AnnotationProvider SPI IMO

impl/src/main/java/org/apache/myfaces/config/annotation/DefaultAnnotationProvider.java
```diff
@@ -128,21 +128,23 @@ public DefaultAnnotationProvider()
     @Override
     public Map<Class<? extends Annotation>, Set<Class<?>>> getAnnotatedClasses(ExternalContext ctx)
     {
-        if (MyfacesConfig.getCurrentInstance(ctx).isUseCdiForAnnotationScanning())
+        //1. Use CDI
+        BeanManager beanManager = CDIUtils.getBeanManager(ctx);
+        if(beanManager != null)  // shouldnt be null, but needed for our mock junit tests
```
volosied, line 133: @tandraschko  Is it okay if I add a null check here?  Otherwise, we'd have to handle the failing `mock` junit tests.
- skip them
- find some way to enable CDI ( looks tedious as we'll need to create lot of mock objects to replicate the CDIAnnotationDelegateInjectionProvider class)

volosied:
As for removing the Annotation Provider, does the quarkus extension need it? I see there's a QuarkusAnnotationProvider.

impl/src/main/java/org/apache/myfaces/config/annotation/DefaultAnnotationProvider.java
tandraschko: yep

tandraschko:
yep, i see
then we should move the CDI logic to DefaultAnnotationProvider and remove the old stuff

volosied:
Could you be more specific -- I don't quite follow which parts you mean?

tandraschko:
I can take care of it If the Mojarra Team give their +1

tandraschko:
i will close this for now and completly remove old scanning after we aligned on the spec change

PR #618, opened 2023-10-03
Spec 1829 : Deprecate full state saving (FSS)

volosied:
for `https://github.com/jakartaee/faces/issues/1829`


melloware (approved):

volosied:
Need to fix the wording in the warning, but is there anything else to be done here?

PR #619, opened 2023-10-03
Spec 1342 : Make Flows Injectable

volosied:
I think this is the right way to do it?

The spec issue also requested conversation scopes to be injectable, but that's done via CDI. 
See https://docs.jboss.org/weld/reference/2.4.0.CR1/en-US/html/scopescontexts.html#_the_conversation_scope ( 5.3.1)


tandraschko:
I think we should produce it in FacesScoped

PR #620, opened 2023-10-06
Fix for MYFACES-4633

Jmcleodfoss:
Look for java:comp/env/BeanManager if request for… java:comp/BeanManager fails rather than if it succeeds


Jmcleodfoss:
I have never done this before - is the intention that I also submit PRs against 4.0.x and 2.3-next?

volosied (approved):

PR #629, opened 2023-10-11
Spec 1263:  Add rowStatePreserved property to UIRepeat 

volosied:
`https://github.com/jakartaee/faces/issues/1263`

I used UIData as a reference so there's a lot of overlap. 5.0 can address the code duplication.


volosied:
I think I'd rather just add the new field rather than make a whole new class. Current UIIterate work will be saved on the `spec1263-uiiterate` branch. I'll update this just to add rowStatePreserved to UIRepeat

tandraschko:
+1

tandraschko:
+1
we need a base class in 5.0

PR #634, opened 2023-10-23
MYFACES-4606 Follow up: do not send values for unselected radio / checkbox elements

volosied:
https://issues.apache.org/jira/browse/MYFACES-4606

### DIFF
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxUtils.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxUtils.js
@@ -60,8 +60,16 @@ _MF_SINGLTN(_PFX_XHR+"_AjaxUtils", _MF_OBJECT,
         if (item && item.type &&
             (item.type.toLowerCase() == "submit" ||
              item.type.toLowerCase() == "button" )) {
+
+            // do not append if unselected
+            if ((item.type == "checkbox" || item.type == "radio") 
+             && issuingItem.attr("checked").isAbsent()) {
+                return;
+            }
+            
             //buttons not always have a name unlike inputs
             targetBuf.append(item.id || item.name, item.value);
+
         }
     },
 


werpu:
Not needed:    if (item && item.type &&
            (item.type.toLowerCase() == "submit" ||
             item.type.toLowerCase() == "button" )) {

The original behavior is somewhat different anyway... i just changed it somehow for the new code to cover more elements (aka every element)
Hence I never have issued a pr for the old code it does not expose the problem!

Either way we theoretically could unify the behavior for all branches, the old code and new code that would be feasible, if wanted, given this is a non standard behavior we are free here to do whatever we want!

volosied:
If I removed the code to just 
```
    appendIssuingItem: function (item, targetBuf) {
        // if triggered by a Button send it along
        if (item && item.type) {

            // do not append if unselected
            if ((item.type == "checkbox" || item.type == "radio") 
             && !item.checked) {
                return;
            }
            
            //buttons not always have a name unlike inputs
            targetBuf.append(item.id || item.name, item.value);
        }
    },
```

then I see the form sent as an array.
Screenshot 2023-10-24 at 3 53 59 PM
Looks like it's caused by the preprocessed data also containing the issuing element? 

Screenshot 2023-10-24 at 3 54 33 PM


Otherwise, with the existing checks, only true is sent: 

    
Screenshot 2023-10-24 at 3 50 50 PM

volosied:
@werpu  What do you recommend here?

werpu:
Yes the original code only allows submits and buttons to be appended as issuing item, I guess we have to replicate the 4.x behavior into the old code might make sense to have a unified behavior, I will prepare something the next few days!

volosied:
@werpu  Any updates? Thank you

volosied:
I would like to merge this if possible soon in order to get new 2.3 3.0 4.0 etc releases out.

werpu:
Hi I will work on the issue today!
Sorry was not aware that it is so urgent.

werpu:
I have issued a pull request for 2.3-next, will also issue one for 3.0, please see: 
https://github.com/apache/myfaces/pull/637

my integration tests pass, it is basically a backport from 4.0

volosied:
Closing in favor of #638.

PR #636, opened 2023-10-27
Fix Myfaces 4623 & Add Test for 5.0.x

volosied:
Fix and test working successfully: 

`[2023-10-27 14:45:02] [autobuild] [INFO] Apache MyFaces Core 5.0 - Integration Tests - myfaces4623 SUCCESS [  7.393 s]`

### DIFF
--- a/integration-tests/myfaces4623/src/test/java/org/apache/myfaces/core/integrationtests/IntegrationTest.java
+++ b/integration-tests/myfaces4623/src/test/java/org/apache/myfaces/core/integrationtests/IntegrationTest.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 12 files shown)


integration-tests/myfaces4623/src/test/java/org/apache/myfaces/core/integrationtests/IntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        trigger("form1:display", webDriver ->
+        {
+            return webDriver.getPageSource().contains("DISPLAYING COMPONENT");
+        });
+        Assert.assertTrue(webDriver.getPageSource().contains("DISPLAYING COMPONENT"));
+
+        trigger("form1:close", webDriver ->
+        {
+            return !webDriver.getPageSource().contains("DISPLAYING COMPONENT");
```
volosied, line 111: This might need to be `DuplicateIdException` instead, but the test still works.

PR #638, opened 2023-11-03
Follow up on https://issues.apache.org/jira/browse/MYFACES-4606

werpu:
Fix for MyFaces 4606, checkboxes and radio buttons are not appended automatically if unchecked!

### DIFF
--- a/api/src/main/javascript/META-INF/resources/myfaces/_impl/_util/_Lang.js
+++ b/api/src/main/javascript/META-INF/resources/myfaces/_impl/_util/_Lang.js
@@ -561,10 +561,16 @@ _MF_SINGLTN(_PFX_UTIL + "_Lang", Object, /** @lends myfaces._impl._util._Lang.pr
         //we simulate the dom level 2 form element here
         var _newCls = null;
         var bufInstance = null;
+        var _Lang = this;
         if (!this.FormDataDecoratorArray) {
             this.FormDataDecoratorArray = function (theFormData) {
                 this._valBuf = theFormData;
                 this._idx = {};
+                var _t = this;
+                _Lang.arrForEach(theFormData, function(item) {
+                    var key = item[0];
+                    _t._idx[decodeURIComponent(key)] = true;
+                });
             };
             _newCls = this.FormDataDecoratorArray;
             _newCls.prototype.append = function (key, val) {
@@ -583,6 +589,12 @@ _MF_SINGLTN(_PFX_UTIL + "_Lang", Object, /** @lends myfaces._impl._util._Lang.pr
                 this._preprocessedData = theFormData;
                 this._valBuf = [];
                 this._idx = {};
+                var _t = this;
+                var keyValuePairs = theFormData.split(/\&/gi);
+                _Lang.arrForEach(keyValuePairs, function(item) {
+                    var key = _Lang.trim(item.split(/\=/gi)[0]);
+                    _t._idx[decodeURIComponent(key)] = true;
+                });
             };
             _newCls = this.FormDataDecoratorString;
             _newCls.prototype.append = function (key, val) {
@@ -591,7 +603,8 @@ _MF_SINGLTN(_PFX_UTIL + "_Lang", Object, /** @lends myfaces._impl._util._Lang.pr
             };
             //for now we check only for keys which are added subsequently otherwise we do not perform any checks
             _newCls.prototype.hasKey = function (key) {
-                return !!this._idx[key];
+                var _t = this;
+                return !!(this._idx[key]);
             };
             _newCls.prototype.makeFinal = function () {
                 if (this._preprocessedData != "") {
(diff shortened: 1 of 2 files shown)


werpu:
Give it a shot, we can merge it, from my side it works!
(tested it with my it test suite)

werpu:
does it work for you?

volosied:
@werpu  I'm getting the following error in my test case: 

```
Error Message: (intermediate value).toLowerCase is not a function

--------------------------------------------------------
Calling class:myfaces._impl.xhrCore._AjaxRequestQuirks
Calling function:send
Error Name: sendError

--------------------------------------------------------
Note, this message is only sent, because project stage is development and no other error listeners are registered. 
```

xhtml:
```
      <h:form id="form1">
         <h:selectBooleanCheckbox id="checkbox" value="#{testBean.checked}">
            <f:ajax render="output"/>
         </h:selectBooleanCheckbox>
         <h:outputText id="output" value = "#{testBean.checked}" />
      </h:form>
      ```

werpu:
mhh let me check quickly!

werpu:
Fixed it, sorry for the issue, I guess I have to recheck my integration tests apparently this code was not triggered by them!
(the 4.x codebase has a dedicated unit test so it is covered there definitely)

volosied:
Yep, it's fixed now. Thanks! 

However, I'm still seeing the an array sent for the issuing element? 

image


Looks like its attached via appendIssuingItem & the  _preprocessedData  (mentioned here: https://github.com/apache/myfaces/pull/634#issuecomment-1777937812)

volosied:
I'm seeing it's caused by 

ajaxUtils.encodeSubmittableFields(ret, form, null);
https://github.com/apache/myfaces/blob/3.0.x/api/src/main/javascript/META-INF/resources/myfaces/_impl/core/Impl.js#L111C19-L111C42

https://github.com/apache/myfaces/blob/3.0.x/api/src/main/javascript/META-INF/resources/myfaces/_impl/xhrCore/_AjaxUtils.js#L38

Can we add some check to avoid encoding the the data twice?

werpu:
Mhh I have to check that let me check that tomorrow!
There needs to be a check that the value of the same element is not attached twice. 
It is getting late here, so I will check that tomorrow!

volosied:
Thanks for looking. I hoped it would work, too, but it doesn't, unfortunately.  Problem looks to be here: 

    getFormData:function () {
        var formDataDecorator = this._Lang.createFormDataDecorator(jsf.getViewState(this._sourceForm)); <-- 
        this._AJAXUTIL.appendIssuingItem(this._source, formDataDecorator);
        return formDataDecorator;
    },
    
    When `jsf.getViewState(this._sourceForm)` is called, it encodes the form into a string.  formDataDecorator's idx variable is then empty, so the new check fails.
    
    Perhaps we would wait to encode things later? 
    
  We can just continue with this more this week. Thanks again for your help here!

werpu:
I will have a deeper look into this tomorrow, I need to debug this out properly, the issue simply is there are possibilities where an array of values can come in over checkbox fields over one id, so I need to perform this check at the appendIssuingItem stage, I will have a deeper look into this tomorrow and will come up with a solution.
The old code simply was not affected by this up until now because only buttons were allowed to be appended, and they were never part of the encoding!
Either way a solvable problem but just not today anymore!

werpu:
Hi, I fixed the check on FormData level, apparently the hasKey function only checked for newly appended values not existing data passed into the existing formData object, this should fix the issue for good, a preliminary test locally now works for me, and all integration tests still pass. I will check whether this issue also is present for the 4.0 code, and if all is well on both code levels we can cherry pick the fixes for the older branches!

volosied:
Hi, the changes look good to me. Both scenarios I've tested pass.

Let me run the full TCK and (hopefully) we merge this in!

werpu:
Any news on the TCK @volosied ?

volosied:
Hi sorry for the delay -- I used the current 3.0.x branch along with your commits to test the TCK. I was hoping to create a 3.0 release soon ( along with the other branches). 

Turns out there's a few failures, but I haven't had a moment  to investigate them. It's possible they aren't related to your changes, but I'll get back to you soon.

melloware:
i thought the 3.0.x branch we decided was dead?

volosied:
We've still been patching fixes, and it's a version our product still supports.    And since the code is mostly the same as 2.3, I expect the TCK failures to also exist in that branch. 

It doesn't take too much time to create a release.

werpu:
I can make a quick pull request for 2.3 (next has one already)

werpu:
Hi i added a small fix from an issue which was reported from the Tobago guys!
The fix for 4.0+ will come soon, it just needs final Tobago testing, but their testcase is fixed by this fix.

werpu:
@volosied there were indeed smaller issues, which the Tobago guys reported my, all of them are now backported to the 3.x branch, please give it another shot!

volosied:
Thanks - I'll test it out again.  Give me a few days!

werpu:
Thanks for testing and closing!

PR #648, opened 2023-11-28
MYFACES-4644 [4.0] Enable extensionless requests for programmatic views

volosied:
fixes https://issues.apache.org/jira/browse/MYFACES-4644


volosied:
I'll see if I can add a test

PR #653, opened 2023-12-08
MYFACES-4646: Fix formatting of string built by ValidatorException

jungm:
Fixes [MYFACES-4646](https://issues.apache.org/jira/browse/MYFACES-4646) by only adding the FacesMessage detail to the ValidatorException message if it is not empty


melloware (approved):

melloware:
Thanks for your fix!  if you want it backported to 4.x and 2.3.x branches please submit PR's against those branches as well!

PR #654, opened 2023-12-08
Create testExtensionlessProgrammaticView for MYFACES-4644

volosied:


integration-tests/pom.xml
```diff
@@ -145,6 +145,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>jakarta.el</groupId>
+            <artifactId>jakarta.el-api</artifactId>
+            <version>5.0.0</version>
```
volosied, line 151: Tried to use EL 6.0.0 M1, but encountered: `` Caused by: java.lang.UnsupportedClassVersionError: jakarta/el/ELContextListener has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0``` 

Using 5.0 to get things to compile for now.

PR #660, opened 2023-12-11
5.0: JDK17 Updates

melloware:
5.0: JDK17 Updates

@tandraschko massive change to update the whole project to JDK17


volosied:
Changes look fine to me -- looks like just Java 17's `Pattern Matching for instanceof` that's refactored here.

Build failed due to a checkstyle error and I see there's a conflict.

melloware:
Let me fix this and resubmit it looks like the checkstyle was re-instated

PR #663, opened 2023-12-14
5.0: JDK17 Adds pattern variables to instanceof expressions

melloware:
Adds pattern variables to instanceof expressions wherever the same (side effect free) expression is referenced in a corresponding type cast expression within the flow scope of the instanceof. Currently, this recipe supports if statements and ternary operator expressions.


tandraschko:
cool, thanks!

PR #667, opened 2023-12-17
MYFACES-4647  fix NPE when the user does not have an active channel

milansie:
Fix NPE when user does not have active channel (ie leave page with connected channel)


melloware:
Do you need to submit one to the 4.0.x branch as well?

milansie:
Yes, it would be great. Or I could prepare PR for 4.0, if it would be easier for you.

melloware:
Yes please submit it for 4.0.x branch!

PR #682, opened 2024-03-14
MYFACES-4655 fix resolving of FacesValidators without value

NicolaIsotta:
https://issues.apache.org/jira/browse/MYFACES-4655

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/cdi/util/CDIUtils.java
+++ b/impl/src/main/java/org/apache/myfaces/cdi/util/CDIUtils.java
@@ -18,21 +18,24 @@
  */
 package org.apache.myfaces.cdi.util;
 
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Type;
-import java.util.Collections;
-import java.util.Set;
 import jakarta.enterprise.context.ContextNotActiveException;
 import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.context.SessionScoped;
 import jakarta.enterprise.context.spi.Context;
-
 import jakarta.enterprise.context.spi.CreationalContext;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.faces.context.ExternalContext;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.view.ViewScoped;
+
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/apache/myfaces/cdi/util/CDIUtils.java
```diff
@@ -120,6 +135,17 @@ public static <T> T get(BeanManager beanManager, Type type, boolean create, Anno
...
 
+    private static String getBeanName(Bean<?> bean)
+    {
+        String name = bean.getName();
+        if (name != null)
+        {
+            return name;
+        }
+        String className = bean.getBeanClass().getSimpleName();
+        return Character.toLowerCase(className.charAt(0)) + className.substring(1);
```
melloware, line 146: These two lines are the only ones I am not sure about but @tandraschko will weigh in!

melloware:
@tandraschko does this need to be backported and forward ported once accepted?

tandraschko:
can you explain the change?

NicolaIsotta:
Not needed, I already tried, here's what happens if you don't declare forClass and value on FacesConverter:
```
@FacesConverter must have value, forClass or both. Check annotation @FacesConverter on class: org.primefaces.test.MyCdiConverter
```

impl/src/main/java/org/apache/myfaces/cdi/util/CDIUtils.java
tandraschko: i also wonder about this one - why is this required?
If bean.getName is null, its not a named bean - why would need it then?

NicolaIsotta: TBH I'm not sure too - guess it depends on how the validator is registered in CDI?

What I'm sure is that without these lines the fix doesn't work...

melloware: It looks like BalusC does this in his CDIUtils for good reason I guess.  he does...

```java
        String className = bean.getBeanClass().getSimpleName();

        return new StringBuilder()
            .append(Character.toLowerCase(className.charAt(0)))
            .append(className.substring(1))
            .toString();
```

tandraschko:
can you provide a PR for 4.1 and 5.0 too?

PR #688, opened 2024-03-27
MYFACES-4657: Fix NPE Caused by null ClassName [4.1]

volosied:
https://issues.apache.org/jira/browse/MYFACES-4657


melloware (approved):

pnicolucci (approved):
Make sure to port this one to main as well.

PR #689, opened 2024-03-27
MYFACES-4657: Fix NPE Caused by null ClassName [5.0]

volosied:


pnicolucci (approved):

tandraschko:
Please make sure to fix in 4.0, 4.1 and 5.0

pnicolucci:
@tandraschko 4.0 doesn't have the code that causes this: https://github.com/apache/myfaces/blob/4.0.x/impl/src/main/java/org/apache/myfaces/webapp/MyFacesContainerInitializer.java#L323 should be all set with 4.1 and 5.0.

PR #701, opened 2024-04-16
Fix up rowStatePreserved in UIRepeat

volosied:
Discovered via TCK testing.

The saveState method wasn't property updated, but it's fixed now.  I used UIData as a reference. 

I added a test case previously, but the TCK is a bit more thorough ( nested ui reeats). 

https://issues.apache.org/jira/browse/MYFACES-4664


melloware (approved):

tandraschko:
Please also for master

tandraschko:
Please dont forget the master

PR #702, opened 2024-04-19
5.0: Quarkus RepeatStatus reflection

melloware:
5.0: Quarkus RepeatStatus reflection, 3.8.4 LTS

This needs to be a reflected as its a POJO and if you reference it in EL for index it fails in Native mode


tandraschko:
@melloware please create a jira issue - no codechanges should be done without

melloware:
got it

PR #711, opened 2024-05-08
MYFACES-4637: Encode Java Records to JSON

volosied:
https://issues.apache.org/jira/browse/MYFACES-4637

Without using a new dependency, this may be the best we can do.

I added a single test case for some basic verification.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/push/Json.java
+++ b/impl/src/main/java/org/apache/myfaces/push/Json.java
@@ -23,6 +23,7 @@ import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Array;
+import java.lang.reflect.RecordComponent;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Collection;
@@ -31,6 +32,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TimeZone;
+import java.util.logging.Logger;
 
 /**
  * A simple JSON encoder.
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/apache/myfaces/push/Json.java
```diff
@@ -111,12 +112,45 @@ else if (object instanceof Class<?>)
...
+            String name = field.getName();
+            builder.append(name);
+            builder.append(':');
+            try {
+                encode(field.get(recordObject), builder);
+            } catch (IllegalArgumentException e) {
+                e.printStackTrace();
+            } catch (IllegalAccessException e) {
+                e.printStackTrace();
+            }
```
volosied, line 143: I'm not sure what I should do with error handling here.

melloware: I feel like we should throw some sort of unchecked exception here right?

impl/src/main/java/org/apache/myfaces/push/Json.java
```diff
@@ -111,12 +112,45 @@ else if (object instanceof Class<?>)
...
     }
 
+    private static void encodeRecord(Object recordObject, StringBuilder builder) {
+
+        builder.append('{');
+        boolean fieldsFound = false;
+
+        for (Field field : recordObject.getClass().getDeclaredFields())
+        {
+            field.setAccessible(true);
```
melloware, line 132: I could be wrong but is this the kind of thing they are removing in JDK 21? Being able to change field access?  I don't follow those JDK reflection updates much but i thought they were removing a lot of things?

melloware (changes requested):

volosied:
@melloware  Take a look at the new code -- I used a Record Specific class to get the fields.

https://download.java.net/java/early_access/panama/docs/api/java.base/java/lang/reflect/RecordComponent.html  ( Only found Java 19 javadoc, but this class was added in 16).

melloware (approved):

melloware:
Slick!

PR #718, opened 2024-06-18
Fixes MYFACES-4671

tssmith04:
Fixes [MYFACES-4671](https://issues.apache.org/jira/browse/MYFACES-4671)
Which is related to the following [OpenLiberty issue](https://github.com/OpenLiberty/open-liberty/issues/27598)
event.data can sometimes be undefined, so this prevents it from being undefined by changing event.data to null in the case when event.data is undefined.

### DIFF
--- a/api/src/client/typescript/faces/impl/PushImpl.ts
+++ b/api/src/client/typescript/faces/impl/PushImpl.ts
@@ -153,7 +153,7 @@ export module PushImpl {
         }
 
         onerror(event: any) {
-            let message = JSON.parse(event.data);
+            let message = JSON.parse(event.data === undefined ? null : event.data);
             //TODO replace this with a more readable Stream code
             for (let i = clientIdsByTokens[this.channelToken].length - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


volosied:
OnMessage also calls parse -- I think we should update it too.

tssmith04:
> OnMessage also calls parse -- I think we should update it too.
One concern I have regarding the comment, if we change `let message = JSON.parse(event.data);` to `let message = JSON.parse(event.data === undefined ? null : event.data);` then message may be null which was acceptable in onerror, but may not be here. Message is used `let functions = behaviors?.[message];` which would throw another error in the case that message was undefined at first because it would become null and an array can't have a null index.
So we will just leave onMessage alone for now.

volosied:
@werpu If this looks good, please merge!

melloware:
we should then forward port to 4.1 and 5.0

werpu:
Hi will will merge it later tonight

werpu:
I will add the fix also to 5.0 later tonight!

PR #720, opened 2024-06-21
Fix MYFACES-4672: G.hasKey is not a function during MultiPart Ajax Requests

volosied:
https://issues.apache.org/jira/browse/MYFACES-4672

Rather than adding hasKey to FormData, I avoid calling `_AJAXUTIL.appendIssuingItem(this._source, ret);`

Code taken from: 

https://github.com/apache/myfaces/pull/415/files#diff-fb08901a91cee0cb1b50e051f32e8c32064936936727f34eea91a9c659f81402R294-R306


volosied:
@werpu  Could you please take a look at this issue? 

I'm not sure if more needs to be done here or if this is enough? Thank you so much.

werpu:
Well this should fix the haskey issue, give it a shot!

werpu:
I would not avoid to call the function, please give https://github.com/apache/myfaces/pull/724 a shot whether it fixes the issue for your customers, if it does then we can merge it in!

PR #724, opened 2024-06-24
https://issues.apache.org/jira/browse/MYFACES-4672:

werpu:
FormData handling not properly working on old code and new browsers!


werpu:
@volosied can you please check, if this fixes your issue?

melloware (approved):

volosied:
This looks good to me, but I have one question -- where is _PartialSubmitUtils used?

werpu:
As far as I remember this is a drop in replacement singleton for the AjaxUtils singleton in the code!
There was this special construct in the old code which allowed to override and delegate singletons. That way I could split seldomly used functionality out codewise into their own classes (did the same with inheritance) to keep a separation of concerns between general implementation and browser specific quirks implementations. Thats also the reason why it was possible relatively quickly to drop a ton of browser specific code in next!

PR #730, opened 2024-07-21
MYFACES-4675: 4.0 ResetValues visit tree

melloware:
https://github.com/jakartaee/faces/issues/1936

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/context/servlet/PartialViewContextImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/context/servlet/PartialViewContextImpl.java
@@ -579,6 +579,11 @@ public class PartialViewContextImpl extends PartialViewContext
             {
                 ((EditableValueHolder)target).resetValue();
             }
+            else if (!VisitContext.ALL_IDS.equals(context.getIdsToVisit()))
+            {
+                // Render ID didn't specifically point an EditableValueHolder. Visit all children as well.
+                target.visitTree(VisitContext.createVisitContext(FacesContext.getCurrentInstance(), null, context.getHints()), this);
+            }
             return VisitResult.ACCEPT;
         }
     }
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/apache/myfaces/context/servlet/PartialViewContextImpl.java
```diff
@@ -579,6 +579,11 @@ public VisitResult visit(VisitContext context, UIComponent target)
             {
                 ((EditableValueHolder)target).resetValue();
             }
+            else if (!VisitContext.ALL_IDS.equals(context.getIdsToVisit()))
+            {
+                // Render ID didn't specifically point an EditableValueHolder. Visit all children as well.
+                target.visitTree(VisitContext.createVisitContext(FacesContext.getCurrentInstance(), null, context.getHints()), this);
```
tandraschko, line 585: can we make the visitContext static or reuseable?
currently this would create tons of the same context object, not quite good from performance perspective
just look at PF one

tandraschko: and please create a unit test for it

melloware: i saw that about the FacesContext but was trying not to make the change too invasive. I will rework this today.

tandraschko: yeah sure
ping if you need help

melloware: OK @tandraschko i might need a hint on this one.  PF does his slighlty differnetly by just calling visitTree where in this case BalusC specifically creates a new VisitContext?

tandraschko:
could you create a unit test first that fails currently?

tandraschko:
see ResetValuesTestCase

melloware:
Failing test case added where it resets `render="@form"`

tandraschko:
@melloware wanna port my commit to 4.1 and 4.0?
no feedback from balusc right now but its the right way, no matter how implemented exactly we need to reset all

melloware:
yep i will look at what you did in 5.0

PR #733, opened 2024-07-22
MYFACES-4675: Failing test case using @form

melloware:


melloware:
@tandraschko shouldn't this have failed without a fix?  It renders="@form".

tandraschko:
I would think so

melloware:
I see the issue that test had `immediate="true"` on it so I need to copy test 2 which is surprisingly completely commented out!

melloware:
OK i uncommented the test and wrote a new test with `@form` but now I see its failing with...

```
 ResetValuesTestCase.testResetValuesActionListenerHandler2:57 » ContextNotActive WebBeans context with scope type annotation @SessionScoped does not exist within current thread
```

tandraschko:
all tests are working if you replace setupRequest with startViewRequest

tandraschko:
and sorry, i dont know why :D the test framework is very old and needs some cleanup actually

melloware:
OK weird now the tests both pass and I expected the new test to fail with @form 🤷🏻‍♂️

tandraschko:
i could image that the tests are not working
could you try to add a negative test?

tandraschko:
your test/post code is not aligned with the xhtml its render=@form, in java code you manually set render to field1, field2

tandraschko:
See https://github.com/apache/myfaces/commit/ceb6509cadbb35e5d6bdb2c5d7a25266a15f7b6e?w=1

PR #734, opened 2024-07-31
MYFACES-4675: 4.0 ResetValues visit tree

melloware:
@tandraschko this is your ported fix from 5.0 but I don't see you doing anything with `PartialViewContextImpl.java` ???

That still just has this:

```java
    // same like UIViewRoot#ResetValuesCallback
    private static class ResetValuesCallback implements VisitCallback
    {
        public VisitResult visit(VisitContext context, UIComponent target)
        {
            if (target instanceof EditableValueHolder)
            {
                ((EditableValueHolder)target).resetValue();
            }
            return VisitResult.ACCEPT;
        }
    }
```

### DIFF
--- a/api/src/main/java/jakarta/faces/component/UIViewRoot.java
+++ b/api/src/main/java/jakarta/faces/component/UIViewRoot.java
@@ -43,9 +43,11 @@ import jakarta.faces.application.ProjectStage;
 import jakarta.faces.application.StateManager;
 import jakarta.faces.component.visit.VisitCallback;
 import jakarta.faces.component.visit.VisitContext;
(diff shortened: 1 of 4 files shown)


tandraschko:
oh
so why the tests are working?

melloware:
Good question?   so does `f:resetValues` different than `p:commandButton resetValues="true"` i know out p:commandButton goes through PartialViewContext?

tandraschko:
ah
PartialViewContextImpl uses UIViewRoot, so it should be ok

melloware:
good for me 😄

PR #736, opened 2024-08-27
MYFACES-4679

volosied:
Draft fix for https://issues.apache.org/jira/browse/MYFACES-4679

This PR assumes only `click` events are submits. Do we have any documentation for this?  Should we check for other events? 

What's the different between `javax.faces.behavior.event` and `javax.faces.partial.event`?  I'm not sure which event should be looked at.

### DIFF
--- a/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlButtonRendererBase.java
+++ b/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlButtonRendererBase.java
@@ -96,6 +96,18 @@ public class HtmlButtonRendererBase
     {
         String clientId = uiComponent.getClientId(facesContext);
         Map paramMap = facesContext.getExternalContext().getRequestParameterMap();
+
+        // MYFACES-4679
+        if(facesContext.getPartialViewContext().isPartialRequest())
+        {
+            String event =  (String) paramMap.get("javax.faces.behavior.event");
+            if("click".equals(event))
+            {
+                return true;
+            }
+            return false;
+        }
+
         FormInfo formInfo = _ComponentUtils.findNestingForm(uiComponent, facesContext);
         String hiddenLink = null;
          
--- a/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlButtonRendererBase.java
+++ b/shared/src/main/java/org/apache/myfaces/shared/renderkit/html/HtmlButtonRendererBase.java
@@ -100,8 +100,7 @@ public class HtmlButtonRendererBase
         // MYFACES-4679
         if(facesContext.getPartialViewContext().isPartialRequest())
         {
-            String event =  (String) paramMap.get("javax.faces.behavior.event");
-            if("click".equals(event))
+            if(HtmlRendererUtils.isPartialOrBehaviorSubmit(facesContext, clientId))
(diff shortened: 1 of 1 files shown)


tandraschko:
all events can be submitted by f:ajax, dont think thats a good fix
also it wasnt ncessary with the old JS codebase?!

volosied:
This fix is only to queue the action event for some scenarios. 

When 4606 was introduced, it allowed all ajax calls (on the form's submit button) to invoke the action event. For example: 

```
    <div>
      <h:inputText />
      <h:commandButton id="confirm3" value="listner3" action="#{ajaxEventBean.confirm()}" >
          <f:ajax event="mouseover"/>
      </h:commandButton>
    </div>
```
The mouseover event would trigger a form submission.  I don't think that's intended. It didn't happen prior to 4606. 


I don't think the current 4606 behavior is correct here, so maybe we should revert it.  However, the question I have is which ajax events should trigger an action event?  I'm having a hard time find what the spec says here. 

Current options: 
1) Revert 4606
2) Only allow certain ajax events to trigger action events. 
3) Other

Although..... re-looking at the fix, the `isSubmitted` method does handle ajax via the `HtmlRendererUtils.isPartialOrBehaviorSubmit(facesContext, clientId);` check in the return.  

 Perhaps some other refactoring could be done here...?  Let me push up a new commit.

tandraschko:
Of course should do trigger an AJAX request:
```
      <h:commandButton id="confirm3" value="listner3" action="#{ajaxEventBean.confirm()}" >
          <f:ajax event="mouseover"/>
      </h:commandButton>
```

volosied:
Yes, an ajax request still occurs, but it's basically a no-op (unless it has a listener attached).  However, should the action be invoked for the mouseover example?

volosied:
The [isPartialOrBehaviorSubmit](https://github.com/apache/myfaces/blob/4.0.x/impl/src/main/java/org/apache/myfaces/renderkit/html/util/HtmlRendererUtils.java#L1058) method is used in the [HtmlLinkRendererBase](https://github.com/apache/myfaces/blob/main/impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlLinkRendererBase.java#L100), [HtmlButtonRendererBase](https://github.com/apache/myfaces/blob/main/impl/src/main/java/org/apache/myfaces/renderkit/html/base/HtmlButtonRendererBase.java#L111), and [HtmlCommandScriptRenderer](https://github.com/apache/myfaces/blob/main/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlCommandScriptRenderer.java#L167). 

By relying on isPartialOrBehaviorSubmit, it would enable some consistent by only allowing "click events" to trigger a form submission.

werpu:
Hi, the fix needs to be done on the client side not the server, I gave an analysis on https://issues.apache.org/jira/browse/MYFACES-4679 also I have issued a pull request https://github.com/apache/myfaces/pull/744 which fixes the issue from the client side without breaking 4606!
Please check it. I don´t think we will have to doctor on the java side to get this running again.

Behavior events should not append the issuing element as key value pair this also was not the intention of 4606 which just tried to replicate the normal submit case patterns in xhr for debugging reasons!

tandraschko:
can we close this now @werpu?

PR #744, opened 2024-09-01
https://issues.apache.org/jira/browse/MYFACES-4679:

werpu:
Fixes for issue


werpu:
This should fix the issue from the js side. The main culprit was that behavior events added the key value pair from the issuing element (4606 patch), and then on the server side an action was triggered automatically

@melloware @tandraschko please test this patch, for me this restores the wanted original behavior!
I guess @tandraschko your merge request now can be dropped if this works well!

volosied:
Thanks for the fix, but I don't think this handles the original scenario in 4606 --

The original test would send the `jakarta.faces.behavior.event	"click"`  as part of the request, and this fix would avoid appending it to the request.

volosied:
Part of me is worried about handing this from the JavaScript side as it would affect all requests while with a Java fix, we would only handle the commandButton behavior.

PR #747, opened 2024-09-04
MYFACES-4666: Overload simpleClassForName -- new param to log exception only

volosied:
https://issues.apache.org/jira/browse/MYFACES-4666

By throwing the faces exception, this logic causes apps to fail during start-up. In our scenario, it happens to the servlet TCK app, as reported in this challenge : `https://github.com/jakartaee/servlet/issues/691` 

This PR tries to find a balance between allowing the app to start up while also alerting the user of the non-existent servlet class. 

The git diff is a bit wonky -- the new simpleClassForName method is the second one, not the first as git shows.

### DIFF
--- a/api/src/main/java/org/apache/myfaces/core/api/shared/lang/ClassUtils.java
+++ b/api/src/main/java/org/apache/myfaces/core/api/shared/lang/ClassUtils.java
@@ -163,14 +163,15 @@ public class ClassUtils
 
     /**
      * Same as {link {@link #simpleClassForName(String)}, but will only
-     * log the exception and rethrow a RunTimeException if logException is true.
+     * log the exception and rethrow a RunTimeException if logAndThrowException is true.
      *
      * @param type
-     * @param logException - true to log/throw FacesException, false to avoid logging/throwing FacesException
+     * @param logAndThrowException - true to log and throw FacesException, false to avoid logging and throwing 
+     *  the FacesException
      * @return the corresponding Class
-     * @throws FacesException if class not found and logException is true
+     * @throws FacesException if class not found and logAndThrowException is true
      */
-    public static Class simpleClassForName(String type, boolean logException)
+    public static Class simpleClassForName(String type, boolean logAndThrowException)
     {
         Class returnClass = null;
         try
@@ -179,7 +180,7 @@ public class ClassUtils
         }
         catch (ClassNotFoundException e)
         {
-            if (logException)
+            if (logAndThrowException)
             {
                 log.log(Level.SEVERE, "Class " + type + " not found", e);
                 throw new FacesException(e);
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/apache/myfaces/util/lang/ClassUtils.java
```diff
@@ -268,6 +269,37 @@ public static Class simpleClassForName(String type, boolean logException)
         return returnClass;
     }
 
+    /**
+     * Same as {link {@link #simpleClassForName(String)}, but accepts two booleans
+     * One to log an exception and another to rethrow a FacesExceptions
+     *
+     * @param type
+     * @param logException - true to log the ClassNotFoundException, false to avoid logging
+     * @param rethrowException - true to throw FacesException, false to avoid logging/throwing FacesException
```
pnicolucci, line 278: False here only avoids throwing the exception.

impl/src/main/java/org/apache/myfaces/util/lang/ClassUtils.java
```diff
@@ -268,6 +269,37 @@ public static Class simpleClassForName(String type, boolean logException)
...
 
+    /**
+     * Same as {link {@link #simpleClassForName(String)}, but accepts two booleans
+     * One to log an exception and another to rethrow a FacesExceptions
+     *
+     * @param type
+     * @param logException - true to log the ClassNotFoundException, false to avoid logging
+     * @param throwException - true to throw a FacesException, false to avoid throwing a FacesException
+     * @return the corresponding Class
+     * @throws FacesException if class not found and logException is true
```
pnicolucci, line 280: throwException is true

api/src/main/java/org/apache/myfaces/core/api/shared/lang/ClassUtils.java
```diff
@@ -163,14 +163,17 @@ public static Class simpleClassForName(String type)
...
      * @param type
-     * @param logException - true to log/throw FacesException, false to avoid logging/throwing FacesException
+     * @param logAndThrowException - true to log and throw FacesException, false to avoid logging and throwing 
+     *  the FacesException
      * @return the corresponding Class
-     * @throws FacesException if class not found and logException is true
+     * @throws FacesException if class not found and logAndThrowException is true
      */
-    public static Class simpleClassForName(String type, boolean logException)
+    // @Override MYFACES-4449: Methods that call ClassUtils.class.getClassLoader() need to be here
```
pnicolucci, line 174: Shouldn't this comment be in the IMPL ClassUtils.java not in the API?

pnicolucci (approved):

melloware (approved):

PR #762, opened 2024-09-30
MYFACES-4683: 4.0 Fix script sort order

melloware:


werpu (approved):
Looks ok to me. I will merge it into the upstream codebase and will add tests there, for a quick myfaces fix this is fine!

PR #769, opened 2024-10-22
https://issues.apache.org/jira/browse/MYFACES-4680:

werpu:
Client side fix, adding the latest js code to the merge request

### DIFF
--- a/api/src/client/package-lock.json
+++ b/api/src/client/package-lock.json
@@ -22,7 +22,7 @@
         "html-webpack-plugin": "^5.5.1",
         "jsdom": "^21.1.1",
         "jsdom-global": "^3.0.2",
-        "jsf.js_next_gen": "4.0.3-beta.3",
+        "jsf.js_next_gen": "4.0.3-beta.5",
         "mocha": "^10.2.0",
         "npm-check-updates": "^16.10.8",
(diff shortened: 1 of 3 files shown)


melloware (approved):

tandraschko:
please dont forget 4.1 and 5.0

volosied (approved):
This fixes the TCK test :) 

thank you for the PR!

volosied:
`myfaces.ab('j_id_c', event, 'renderAjaxOutput', '', 'ajaxOutput')`  -- maybe a future revision could look for form where the render argument (ID value) is located? 

But I think this change is good enough for this test (and no users have reported this so far).

werpu:
That was not the problem if the TCK test, the problem was that the render ID was outside of any form, but given there is exactly one form in the page it was safe enough to use that one for the viewstate handling!
It still would fail with a two form setup, but then you could use a form passthrough parameter to reference the form you want to use (a feature Mojarra probably does not have in its js implementation, but perfectly html5 compliant)

werpu:
I will start take care of https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4685 tomorrow, I want to wrap up the other branches with the websocket fix today!

PR #805, opened 2024-11-20
MYFACES-4654: Support Schema Validation for 4.1

volosied:
Pulled from https://jakarta.ee/xml/ns/jakartaee/

1)  web services client xsd was pulled in to fix this error: 
```
jakarta.faces.FacesException: org.xml.sax.SAXParseException; systemId: file:... jakartaee_11.xsd; lineNumber: 106; columnNumber: 52; src-resolve: Cannot resolve the name ‘jakartaee:service-refGroup’ to a(n) ‘group’ component.
```

2) standard config xml was updated to 2.3 because `<client-window-factory>org.apache.myfaces.lifecycle.clientwindow.ClientWindowFactoryImpl</client-window-factory>` is used. It was added in 2.3. OTherwise validation failures for 2.0 to 2.2. 

3) Added support for XML validation for 2.2-4.1

4) DTD for 1.0 still works. Verifed via 
```
<?xml version="1.0"?>
<!DOCTYPE facelet-taglib PUBLIC
"-//Sun Microsystems, Inc.//DTD Facelet Taglib 1.0//EN"
"https://java.sun.com/dtd/facelet-taglib_1_0.dtd">
<facelet-taglib>
...
```

5) This change will be ported to earlier versions, but later on. I'll keep the JIRA open for now.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java
+++ b/impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java
@@ -60,7 +60,10 @@ public class ConfigFilesXmlValidationUtils
     private final static String FACES_CONFIG_SCHEMA_PATH_23 = "org/apache/myfaces/resource/web-facesconfig_2_3.xsd";
     private final static String FACES_CONFIG_SCHEMA_PATH_30 = "org/apache/myfaces/resource/web-facesconfig_3_0.xsd";
     private final static String FACES_CONFIG_SCHEMA_PATH_40 = "org/apache/myfaces/resource/web-facesconfig_4_0.xsd";
-    private final static String FACES_TAGLIB_SCHEMA_PATH = "org/apache/myfaces/resource/web-facelettaglibrary_2_0.xsd";
+    private final static String FACES_TAGLIB_SCHEMA_PATH_20 = 
+                                                        "org/apache/myfaces/resource/web-facelettaglibrary_2_0.xsd";
+    private final static String FACES_TAGLIB_SCHEMA_PATH_41 = 
+                                                        "org/apache/myfaces/resource/web-facelettaglibrary_4_1.xsd";
 
     public static class LSInputImpl implements LSInput
     {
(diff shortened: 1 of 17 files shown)


volosied:
@bohmber @tandraschko Can I pull in these xsd files? My concern is the copyrights. 

` SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0`

volosied:
I get the get the feeling it's a no -

```
This is a copyleft license, meaning that any derivative works created using GPLv2 code must also be released under the GPLv2, which can be incompatible with the more flexible Apache license.
```

volosied:
But tomcat also has it? https://github.com/apache/tomcat/blob/main/java/jakarta/servlet/resources/jakartaee_11.xsd

I'm honestly not sure.

melloware:
It is very weird these are GPL2?

bohmber:
[ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html)
[EPL 2.0](https://www.eclipse.org/legal/epl-2.0/) is Category B only the object/binary form is allowed. But there is an exception: 

`For small amounts of source code that the ASF product directly consumes at runtime, and for which that source is unmodified and unlikely to be changed anyway (say, by virtue of being specified by a standard), you may include appropriately labeled source code. An example of this is the web-facesconfig_1_0.dtd, whose inclusion is mandated by the JSR 127: JavaServer Faces specification.`

melloware (approved):

melloware:
No objections here

impl/src/main/java/org/apache/myfaces/config/ConfigFilesXmlValidationUtils.java
```diff
@@ -530,44 +584,61 @@ public static void validateFaceletTagLibFile(URL xmlFile, ExternalContext extern
...
         }
-        else
+
+        if (stream == null)
         {
-            return "1.0";
+            return null;
         }
+
+        System.out.println("USING: "  + tagLibraryPath);
```
volosied, line 628: Remove

pnicolucci (approved):

bohmber:
@volosied all these new files are CDDL or EPL-2.0?

bohmber (changes requested):
All these new files are CDDL or EPL-2.0?

volosied:
I retrieved them from `http://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/index.html` 

Looking at the documentation within: 
```
The contents of this file are subject to the terms of either the GNU
      General Public License Version 2 only ("GPL") or the Common Development
      and Distribution License("CDDL")
```

volosied:
So they are both.  Some are EPL 2.0 (for Jakarta Files) while the older ones (Javax Files) are CDDL+GPL_1_1

bohmber (approved):

PR #806, opened 2024-11-20
MYFACES-4695: Allow searching for ids with separator character [5.0]

volosied:


volosied:
Added a test -- without my fix it fails with 
```
[ERROR] Errors: 
[ERROR]   SearchExpressionImplTest.testMyFaces4695:423 » ComponentNotFound Cannot find component for expression "form1:one" referenced from "form2:submit".
[INFO] 
```

tandraschko:
i dont think the case is valid, it needs to be
`<f:ajax render=":form1:one" />`

volosied:
I see this as more of a convenience. 

The spec describes the search algorithm here: https://jakarta.ee/specifications/faces/4.1/apidocs/jakarta.faces/jakarta/faces/component/search/searchexpressionhandler

However, it also states "The search algorithm must operate as follows, **though alternate algorithms may be used as long as the end result is the same:**"

If the component can't be found, then it will throw an exception, but, in my opinion, it should keep searching, if possible to find the component.

As for the code, we do a bottom - up search (see this [line](https://github.com/apache/myfaces/blob/173739656b99a06b67aa493084db5c2a41c1f4cf/impl/src/main/java/org/apache/myfaces/component/search/SearchExpressionHandlerImpl.java#L535C1-L535C70))  

if we can't find the component, should we use try from the root instead? 

Let me know what you think. Thanks

tandraschko:
its clearly described in the specs:
`Otherwise, search up the parents of this component. If a [NamingContainer](https://docs.oracle.com/javaee/7/api/javax/faces/component/NamingContainer.html) is encountered, it will be the base.`

https://docs.oracle.com/javaee/7/api/javax/faces/component/UIComponentBase.html#findComponent-java.lang.String-

tandraschko:
and its like this for ages
either we dont support it or we add a config, disabled per default, to allow this

volosied:
One more data point -- this previously worked in 2.2, but then changed in 2.3 due to addition of the SearchExpressionHandler API.

tandraschko:
i think it was a hack around it, it matches your example:
https://github.com/apache/myfaces/blob/2.2.x/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlAjaxBehaviorRenderer.java#L498

tandraschko:
im really happy that both myfaces and mojarra uses almost the exact same impl for this very old topic and that we could even remove the native PF expressions

if add such a hack again, we should make it disabled per default, just to allow users to migrate their old app
but still then... its not really a big change to fix their expressions

tandraschko:
Can you Test it in mojarra?

volosied:
Mojarra 4.1.2 works without needing the prepended `:` -- it find the component.

tandraschko:
Then let me check it
In mojarra they use a invokeOnComponent Fallback, Leo optimized the algorithm in mf

impl/src/main/java/org/apache/myfaces/component/search/SearchExpressionHandlerImpl.java
```diff
@@ -534,8 +534,8 @@ public void invokeContextCallback(FacesContext context, UIComponent target)
             // At this point if the algorithm hasn't returned and the topExpression does not have any separator char
             // we need to do the search backward using findComponent.
             if (target == null 
-                    && searchExpressionContext.getSource() == previous 
-                    && (topExpression.indexOf(separatorChar) == -1) )
+                    && searchExpressionContext.getSource() == previous )
+                    // && (topExpression.indexOf(separatorChar) == -1) ) removed for MYFACES-4695
```
tandraschko, line 538: if you replace it with 
`&& (topExpression.indexOf(":@") == -1))`

im fine with it; please dont forget all other branches

tandraschko:
+1
just proceed

PR #814, opened 2024-12-05
MYFACES-4696: 5.0 Undertow connection abort handling

melloware:
@tandraschko here is what I came up with please review


melloware:
I have never gotten an error before I wasn't sure how to reproduce it?  If you give me a hint i can try it.

tandraschko:
i have a quarkus project primefaces poseidon
sometimes after restart, and reload page i get this
its propably only occurs in firefox as it loads both from cache and from network -> if cache is faster and available, network connection is aborted

melloware:
Weird I am trying in in Firefox and I see those NS_BINDING_ABORTED but nothing in my logs.

image

tandraschko:
cant replicate it currently (think it will soon there again) but its safe to merge at least for 5.0

PR #828, opened 2025-01-03
[2.2] MYFACES-4705

volosied:
https://issues.apache.org/jira/browse/MYFACES-4705

[MYFACES-4500](https://issues.apache.org/jira/browse/MYFACES-4500) and [MYFACES-4524](https://issues.apache.org/jira/browse/MYFACES-4524) were not ported to 2.2.

[MYFACES-4559](https://issues.apache.org/jira/browse/MYFACES-4559) does not apply to 2.2


tandraschko:
are sure anyway to still commit in 2.2?
i would like it to see it EOL

tandraschko:
same as 3.0 actually...
2.3 as only maintained javax branch
4.0, 4.1, 5.0 as currently only maintained jakarta branch

volosied:
I still would like to get 3.0.3 out as we have users on it.  (for those who are close to migrate to 4.0, but would like Jakarta namespace)

PR #832, opened 2025-01-13
MYFACES-4708: Quarkus 3.8.2

melloware:


volosied:
Still failing for me: 

```
[ERROR] myfaces/extensions/quarkus/runtime/src/main/java/org/apache/myfaces/core/extensions/quarkus/runtime/scopes/QuarkusClientWindowScopedContext.java:[21,22] cannot access io.quarkus.arc.InjectableContext
[ERROR]   bad class file: .m2/repository/io/quarkus/arc/arc/3.8.2/arc-3.8.2.jar(/io/quarkus/arc/InjectableContext.class)
[ERROR]     class file has wrong version 61.0, should be 55.0
```

Looks like Java 17 was introduced in 3.7: https://quarkus.io/blog/java-17/ 


Alternatively, I wonder if you can set a higher java level for the quarkus sub-projects? However, that means  Faces standalone needs JDK 11, but with Quarkus will require Java 17.

melloware:
ahhh damn.

melloware:
I bumped it to 17 but i have no idea if thta will work just for the extension?

volosied:
*facepalm*  You're right -- my suggestion was bad....  since JDK 11 doesn't know about 17. 

Any chance you can use an older version?  I verified that 3.6.9 still uses Java 11:

```
javap -v  io/quarkus/arc/runtime/devconsole/InvocationsMonitor.class  | grep major
  major version: 55
```

Can we recommend any users who need quarkus 3.7 or higher use Faces 4.1 instead?   It's a minor release, so it shouldn't pose any migration issues.

volosied:
Or we change the required JDK to Java 17 for 4.0 -- target Java 11 for the API / IMPL but quarkus for 17.  I think this would need some community discussion.  

Alternatively, we just keep 3.15.1, if possible?  I don't know too much about these different quarkus libraries, but we don't change the compilation level for MyFaces.

melloware:
yep let me back up to 3.6.9 no problem. 4.0 should stay on 11 that is my bad!

melloware:
I reverted back to 3.2.9 which is the last LTS release before 3.8

PR #833, opened 2025-01-21
Faces 1713: Step Backwards

melloware:
Draft for https://github.com/jakartaee/faces/issues/1713

It looks like half of this code used to support backward steps?


melloware:
i have to come back to this, this was not as easy as i thought it was going to be

PR #837, opened 2025-02-06
MYFACES-4709: Add FacesConfig to HandlesTypes

volosied:
As described in the comment: 

```
 * ... the listener checks if 
 * the FacesServlet has already been defined in web.xml and if not, it adds
 * the FacesServlet with the mappings (/faces/*, *.jsf, *.faces) dynamically.
```

In summary, if this annotation is found and the web.xml doesn't have any mappings listed for the faces servlet, they will be added automatically. 

I'm assuming this was mise when this annotation was added 2.3


melloware (approved):

melloware:
Does this need to be backported?

volosied:
Yes, I'll do that soon.

melloware:
Thank you!

PR #848, opened 2025-02-26
https://issues.apache.org/jira/browse/MYFACES-4710: Fix for swallowed parameters in commandscript

werpu:
This fixes the issue described in https://issues.apache.org/jira/browse/MYFACES-4710!
Please review the changes. I had to add an extenions in the AjaxScriptRenderer, and need additional eyes to review the changes.
The idea is that I have extended the myfaces.ab function with user parameters which allow additional parameters to be added which then later are properly merged in. This however required some code changes on the java side, because the original issue was that the "o" parameter was never passed in in the generated javascript!
Note for "speed" reasons I just dragged the typescript changes in, I will roll a proper script release once this bug is fixed because I need to add some typo fixes done on the myfaces side, back into the upstream project!

### DIFF
--- a/api/src/client/typescript/faces/@types/definitions/index.d.ts
+++ b/api/src/client/typescript/faces/@types/definitions/index.d.ts
@@ -125,7 +125,7 @@ declare global {
     }
 
     interface MyFacesAPI {
-        ab(source: Element, event: Event, eventName: string, execute: string, render: string, options: Context): void;
+        ab(source: Element, event: Event, eventName: string, execute: string, render: string, options?: Context, userParams?: Context): void;
 
         config: { [key: string]: any };
         oam: OAM;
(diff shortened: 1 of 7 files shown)


melloware (approved):

melloware:
looks like one failing test

werpu:
yes I noticed and fixed the failing test immediately, it was that the test was not valid anymore due to parameters being slightly changed (one added) in the AjaxScriptRenderer!

melloware:
looks good to me @tandraschko any input before we merge and backport?

werpu:
the o (name is not ideal) is a set of user parameters which can be passed as per commandScript spec!
basically you can define a function via command and the user can pass a set of parameters which are then passed as ajax request parameters down the request chain!
The initial problem was that this o parameter was swallowed. We really should rename it but given this drags into the AjaxScriptRenderer I am not eager to do this in this bugreport!

tandraschko:
can you just add a comment here after the "o" with something like "see CommandScriptRenderer someMethod", so we know that its not just a random "o" ;D

Then +1 to merge

werpu:
Sure I will... makes sense, I will add it later today and then I will proceed with the merge!
Thanks for the rewiew and input!

PR #863, opened 2025-03-17
[4.1.x] Avoid NPE in logException's while loop

volosied:
1) I split the loop up into two parts for readability/simplicity.  One for the FacesWrapper and the other for the FacesException/ELException.
2) Added a check for null. If so,  we exit the loop.


volosied:
Testing is still in progress.

tandraschko:
Looks good

PR #869, opened 2025-03-19
Add null check for contentTypeListString

volosied:
The JavaDoc for this method:
https://docs.oracle.com/javaee/7/api/javax/faces/render/RenderKit.html#createResponseWriter-java.io.Writer-java.lang.String-java.lang.String-

Key Points: 
`contentTypeList - an "Accept header style" list of content types for this response, or null if the RenderKit should choose the best fit. `

`[IllegalArgumentException](http://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html?is-external=true) - if no matching content type can be found in contentTypeList, no appropriate content type can be found with the implementation dependent best fit algorithm, or no matching character encoding can be found for the argument characterEncoding.` 
_______ 

The  `contentTypeListString` is a parameter. If it is null, then we should should the best fit (i.e the default as in our case) as described in the javadoc. 

If it's not null, then myfaces should continue to try to find a match as on line 289. (If `*/*` is Accepted, then use the default. Otherwise, if `*/*` is not accepted, then the IllegalArgumentException is thrown since not match was found)

I don't like this code, but it looks to work for me locally to pass the TCK test while still avoid the infinite loop.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlRenderKitImpl.java
+++ b/impl/src/main/java/org/apache/myfaces/renderkit/html/HtmlRenderKitImpl.java
@@ -280,11 +280,18 @@ public class HtmlRenderKitImpl extends RenderKit implements LazyRenderKit
                 {
                     if (contentTypeListStringFromAccept != null)
                     {
-                        String[] contentTypes = ContentTypeUtils.splitContentTypeListString(
-                                contentTypeListStringFromAccept);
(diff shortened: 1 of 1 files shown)


volosied:
@melloware  when you have a moment, can you also test to ensure the infinite loop is avoided?

melloware (approved):
Yes this seems to fix it!

PR #872, opened 2025-03-20
[5.0.x] Add serverError prefix to comply with TCK

volosied:


volosied:
@werpu  Hope this isn't an issue with you as it's a fairly small change. I'll can update this later if necessary, but it passes the TCK test.

PR #879, opened 2025-04-21
[2.3.x] Add one-time-code support to AUTOCOMPLETE_OFF_VIEW_STATE

volosied:
I made the _autoCompleteOffViewState variable a String instead to support `one-time-code`. 

Also changed the default value back to true (as it before prior).

This also presents an issue where AUTOCOMPLETE_OFF_VIEW_STATE isn't an appropriate name anymore since it supports a new value. 

However, to avoid regressions or adding new context parameters, I think we can let this pass?


melloware (approved):

volosied:
@pnicolucci @melloware  Please take another look.  I refactored this a bit more and tested to ensure it works as expected. 

- **param value - output** 
- true - autocomplete=off
- false - no autocomplete
- one-time-code - autocomplete=one-time-code
- invalid config -  autocomplete=off

melloware (approved):

pnicolucci (approved):
Thanks Vlad this seems reasonable!

PR #881, opened 2025-04-22
[5.0] Myfaces 4721 - AUTOCOMPLETE_VIEW_STATE_VALUE

volosied:
Draft

New Parameter: org.apache.myfaces.AUTOCOMPLETE_VIEW_STATE_VALUE

Values:
- disable (not autocomplete attribute is written).  (1)
- off (backwards compatibility)  
- one-time-code (default)

1) Note `If an <input>, <select> or <textarea> element has no autocomplete attribute, the browser will use the autocomplete attribute of the element's owning form. `  ([source](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#description))


tandraschko:
will close it for now

PR #900, opened 2025-06-06
Fix MYFACES-4719 Spacing

volosied:
https://issues.apache.org/jira/browse/MYFACES-4719

One of the issues we found, but there are few others left in this JIRA. 

This matches the 4.0+ code: https://github.com/apache/myfaces/blob/4.0.x/impl/src/main/java/org/apache/myfaces/renderkit/html/util/AjaxScriptBuilder.java#L302


volosied:
Closing in favor of #901

PR #902, opened 2025-06-16
MYFACES-4706: Sync JavaScript beween 2.3 and 3.0

volosied:
There aren't that many changes between 2.3 and 3.0 related to the namespace, so I think it makes sense to just copy our latest 2.3.x and have ported to 3.0.x.

This keeps our code in sync and easier to maintain, so special edge cases / scenarios for 3.0.x. It will be one code-base used between the two EE levels.


melloware (approved):

volosied:
there's java code that needs to be updated, too...

tandraschko:
thats something @werpu should check too

PR #911, opened 2025-06-24
MYFACES-4726 [4.1.x] Use SHA256DRBG,DRBG,SHA1PRNG as default psuedo-random generator

volosied:
This is considered a breaking change (as 4.1 is released now), so we don't have to merge it. I'll let the community decide. 

There is value in updating to a stronger SHA algorithm. 

https://issues.apache.org/jira/browse/MYFACES-4726

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/util/token/SessionIdGenerator.java
+++ b/impl/src/main/java/org/apache/myfaces/util/token/SessionIdGenerator.java
@@ -57,21 +57,25 @@ public class SessionIdGenerator
     
     /**
      * The name of the algorithm to use to create instances of
-     * {@link SecureRandom} which are used to generate session IDs. If no
-     * algorithm is specified, SHA1PRNG is used. To use the platform default
-     * (which may be SHA1PRNG), specify the empty string. If an invalid
-     * algorithm and/or provider is specified the {@link SecureRandom} instances
-     * will be created using the defaults. If that fails, the {@link
+     * {@link SecureRandom} which are used to generate session IDs.
+     * 
+     * If an invalid algorithm and/or provider is specified the {@link SecureRandom}
+     * instances will be created using the defaults. If that fails, the {@link
      * SecureRandom} instances will be created using platform defaults.
+     * 
+     * The "SHA256DRBG, DRBG, and SHA1PRNG" options were introduced in 4.1.2.
+     * These are orderd by priority.  
+     * The MyFacesConfig#selectAvailableSecureRandomAlgorithm determines
+     * which is available. 
      */
-    private String secureRandomAlgorithm = "SHA1PRNG";
+    private String secureRandomAlgorithm = null;
     
     /**
      * The name of the provider to use to create instances of
      * {@link SecureRandom} which are used to generate session IDs. If no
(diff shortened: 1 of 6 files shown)


melloware:
personally i am OK with it

volosied:
@tandraschko  Your thoughts?  Should we fix it in 4.1 (since it's recently released) or just in 5.0 instead?

tandraschko:
i dont see a reason to not update 4.0,4.1,5.0 to the same

impl/src/main/java/org/apache/myfaces/util/token/TokenGenerator.java
```diff
@@ -31,23 +34,46 @@ public class TokenGenerator
...
     public TokenGenerator()
     {
         seed = new AtomicLong(generateSeed());
     }
     
     private long generateSeed()
     {
-        SecureRandom rng;
-        try
+        SecureRandom rng = null;
```
tandraschko, line 51: i assume generateSeed will be called again and again
so please cache rng - always re-lookup it causes exceptions and exceptions each time

volosied: It's only called within the constructor. 

generateSeed is only called once per instance but I don't thin there should be harm in making it static and caching it.

tandraschko: then we should rework this completely

we should look all allgorithm on startup (MyFacesConfig?) and only remember the algorithm to use

TokenGenerator/SessionIdGenerator should not try and try over and over

impl/src/main/java/org/apache/myfaces/util/token/SessionIdGenerator.java
```diff
@@ -18,35 +18,19 @@
...
-     * designed this way since random number generators use a sync to make them
-     * thread-safe and the sync makes using a a single object slow(er).
+{   
+    /*
+     *  For performance, isolating SecureRandom between threads. 
      */
-    private Queue<SecureRandom> randoms = new ConcurrentLinkedQueue<>();
-    
+    private static ThreadLocal<SecureRandom>  randomLocal = new ThreadLocal<>();
+
```
volosied, line 33: I removed the queue as I don't think it's necessary anymore? 

SecureRandom has been thread safe for years now (since Java 7).  However, it's probaly not best to reuse the secure random instance with multiple threads, but, if we make it thread local, then I think it should be safe? 

I shouldn't expect any performance hits, at least. 

Please review.

volosied: I use a static instance, so it's still reused. ThreadLocal doesn't make any impact here..

Need to review this some more....

impl/src/main/java/org/apache/myfaces/util/token/TokenGenerator.java
tandraschko: ping me if you did it

volosied:
SHA256DRBG is not available on all JDKs. I don' think it should be the default.

SHA1PRNG is still used by Tomcat, so I think there's no rush to move to something stronger just yet. 

Closing this for now, but perhaps we can revisit it.

PR #912, opened 2025-06-24
MYFACES-4726 [5.0.x] Use SHA256DRBG as default psuedo-random generator

volosied:
Setting as the default for 5.0

https://issues.apache.org/jira/browse/MYFACES-4726


volosied:
Conflicts...? I'll have to look at them.

volosied:
SHA256DRBG is not available on all JDKs. I don' think it should be the default.

SHA1PRNG is still used by Tomcat, so I think there's no rush to move to something stronger just yet. 

Closing this for now, but perhaps we can revisit it.

PR #914, opened 2025-06-25
Fix end of files; files should end with one newline

jbampton:


bohmber:
@jbampton did you test the result?

PR #927, opened 2025-09-03
MYFACES-4731

volosied:
This PR handles a few things
- Syncs up with main (5.0) via Thomas' EL commit.
- Updates dependencies and exports to align with EE11
- Introduces DISABLE_OPTIONAL_EL_RESOLVER

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
+++ b/impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
@@ -785,6 +785,15 @@ public class MyfacesConfig
     public static final String EXCEPTION_TYPES_TO_IGNORE_IN_LOGGING =
             "org.apache.myfaces.EXCEPTION_TYPES_TO_IGNORE_IN_LOGGING";
 
+
+    /**
+     * Expression Language 6.0's OptionalELResolver is enabled by default, so this option allows us to disable it.
+     */
+    @JSFWebConfigParam(name="org.apache.myfaces.DISABLE_OPTIONAL_EL_RESOLVER", since="4.1.2", defaultValue = "false")
+    public static final String DISABLE_OPTIONAL_EL_RESOLVER = "org.apache.myfaces.DISABLE_OPTIONAL_EL_RESOLVER";
+    public static final boolean DISABLE_OPTIONAL_EL_RESOLVER_DEFAULT = false;
+
+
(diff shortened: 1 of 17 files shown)


tandraschko:
no, actually like always

if (el6) {addNew(OptionalELResolver.class)}

and you can just comment the overwrite on getFeatureDescriptors, like we did with BeanValidation methods in the past

impl/src/main/java/org/apache/myfaces/config/webparameters/MyfacesConfig.java
```diff
@@ -1816,5 +1828,11 @@ public List<String> getExceptionTypesToIgnoreInLogging()
     {
         return exceptionTypesToIgnoreInLogging;
     }
+
+    public boolean isOptionalELResolverDisabled()
+    {
+        System.out.println("returning: " + disableOptionalResolver);
```
volosied, line 1834: Remove

pnicolucci, line 1834: remove system.out

tandraschko:
actually there is no big difference
4.1 was EL5 compatible, so we should not break it. This means that we dont remove `getFeatureDescriptors` but just remove/comment `@Override` from it

and it should work on both EL5 and EL6, right?

If we wanna have EL5 compile dependency, we can lookup OptionalELResolver via reflection and it add
Or we use EL6 as dependency add `if (el6)` around the `.add(new OptionalELResolver())`

volosied:
To include the OptionalELResolver and ELRecordResolver, we need to use EL 6.0 dependency. 

However, getFeatureDescriptors was not the only item removed in 6.0 Both ELResolver.TYPE and ELResolver.RESOLVABLE_AT_DESIGN_TIME were also taken out.

Only way I can keep compatibility is to remove the `@Overrides` and hard code the RESOLVABLE_AT_DESIGN_TIME / TYPE values. 

However, CopyBeanInterceptorELResolver is one location that directly calls getFeatureDescriptors.  I'll have to use reflection? 

I've pushed my draft up, but will need to run more tests.

volosied:
I'm unable to test on EL 5.0, but take a look.   Everything does also compile with Tomcat EL 10.1 (if I remove the references to the new EL Resolvers)

volosied:
@tandraschko  I'll merge this if you don't have any more comments. Thanks!

PR #987, opened 2026-05-03
MYFACES-4749 4.0.x branch

werpu:
See https://issues.apache.org/jira/browse/MYFACES-4749 for a changelist


melloware:
Hmmm not sure we should backport this to 4.0.X but i will leave that up the @volosied @tandraschko

werpu:
Actually i do not want to have a code split, up until now the 3 codebases were the same on all 3 branches. Thing is still maintaining the 2.x codebase parallelly to the new one is burden enough, having another split in the new codebase is not a good idea! But as usual, this is a common decision within the project!

PS:
The only reason I would see for a freeze or split would be if we would move the new codebase to a higher EcmaScript standard than what we support now, but even Typescript 7 will have Es2015 as lowest baseline so we are save in this regard until TS decides to cut off ES2015 support!

melloware:
Totally understood.

PR #1022, opened 2026-06-19
MYFACES-4757: 5.0 Fix race conditions of CopyOnWriteList

melloware:
@tandraschko smallest possible fix I could make is that that the problem is that `CopyOnWriteArrayList.size()` and `get()` are not atomic so by using the `for each` the collection is only grabbed once.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
+++ b/impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
@@ -123,7 +123,7 @@ public class ApplicationImplEventManager
         Assert.notNull(listener, "listener");
 
         // MYFACES-4757 use computeIfAbsent to prevent multiple threads from creating different lists
-        List eventInfos = globalListeners.get(systemEventClass);
+        List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
         if (eventInfos == null)
         {
             eventInfos = globalListeners.computeIfAbsent(systemEventClass, k -> new CopyOnWriteArrayList<>());
--- a/impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
+++ b/impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
@@ -122,11 +122,11 @@ public class ApplicationImplEventManager
         Assert.notNull(systemEventClass, "systemEventClass");
         Assert.notNull(listener, "listener");
 
-        List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
+        // MYFACES-4757 use computeIfAbsent to prevent multiple threads from creating different lists
+        List eventInfos = globalListeners.get(systemEventClass);
         if (eventInfos == null)
         {
-            eventInfos = new CopyOnWriteArrayList<>();
-            globalListeners.put(systemEventClass, eventInfos);
+            eventInfos = globalListeners.computeIfAbsent(systemEventClass, k -> new CopyOnWriteArrayList<>());
         }
 
         EventInfo eventInfo = new EventInfo();
@@ -239,9 +239,9 @@ public class ApplicationImplEventManager
             return event;
         }
 
-        for (int i  = 0, size = listeners.size(); i < size; i++)
+        // MYFACES-4757: Use for each to prevent race condition
+        for (SystemEventListener listener : listeners)
         {
-            SystemEventListener listener = listeners.get(i);
             if (listener.isListenerForSource(source))
             {
                 // Lazy construct the event; zhis same event instance must be passed to all listener instances.
@@ -353,9 +353,9 @@ public class ApplicationImplEventManager
             return event;
         }
         
-        for (int i  = 0, size = eventInfos.size(); i < size; i++)
+        // MYFACES-4757: Use for each to prevent race condition
+        for (EventInfo eventInfo : eventInfos)
         {
-            EventInfo eventInfo = eventInfos.get(i);
             if (eventInfo.sourceClass != null && !eventInfo.sourceClass.isAssignableFrom(sourceBaseType))
             {
                 continue;


tandraschko:
is it only broken on master?

melloware:
No I will backport but wanted you to approve it first.

impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
```diff
@@ -122,12 +122,8 @@ public void subscribeToEvent(Class<? extends SystemEvent> systemEventClass, Clas
         Assert.notNull(systemEventClass, "systemEventClass");
         Assert.notNull(listener, "listener");
 
-        List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
-        if (eventInfos == null)
-        {
-            eventInfos = new CopyOnWriteArrayList<>();
-            globalListeners.put(systemEventClass, eventInfos);
-        }
+       List<EventInfo> eventInfos =
```
tandraschko: the old way is faster

tandraschko:
and please add a comment like: Use for-each loop because the list can be changed during calling event listeners

ohterwise i will auto-refactor next time

melloware:
here was the note from AI on computeIfAbstent

Yes, this is a race condition.

The bug is exactly here:

```java
for (int i = 0, size = eventInfos.size(); i < size; i++)
{
    EventInfo eventInfo = eventInfos.get(i);
    ...
}
```

`eventInfos` is actually a `CopyOnWriteArrayList<EventInfo>`:

```java
eventInfos = new CopyOnWriteArrayList<>();
```

The problem is that `CopyOnWriteArrayList.size()` and `get()` are **not atomic**. Consider:

```
Thread A                         Thread B
--------                         --------

size = eventInfos.size();   // 3

                                 unsubscribeFromEvent()
                                 removeIf(...)
                                 backing array now size = 2

eventInfos.get(2);          // ArrayIndexOutOfBoundsException
```

`CopyOnWriteArrayList` intentionally replaces its internal array on every modification. The `size()` came from one snapshot while `get()` reads from a newer snapshot.

This is exactly the exception you're seeing:

```
Index 2 out of bounds for length 2
```

---

## The correct fix

The JDK documentation recommends iterating using the snapshot iterator.

Instead of

```java
for (int i = 0, size = eventInfos.size(); i < size; i++)
{
    EventInfo eventInfo = eventInfos.get(i);
    ...
}
```

use

```java
for (EventInfo eventInfo : eventInfos)
{
    if (eventInfo.sourceClass != null &&
        !eventInfo.sourceClass.isAssignableFrom(sourceBaseType))
    {
        continue;
    }

    if (eventInfo.listener.isListenerForSource(source))
    {
        if (event == null)
        {
            event = createEvent(systemEventClass, facesContext, source);
        }

        if (event.isAppropriateListener(eventInfo.listener))
        {
            event.processListener(eventInfo.listener);
        }
    }
}
```

or equivalently

```java
for (Iterator<EventInfo> it = eventInfos.iterator(); it.hasNext();)
{
    EventInfo eventInfo = it.next();
    ...
}
```

The iterator captures the backing array once, so it is immune to concurrent additions/removals.

---

## There are other occurrences in this class

I also noticed `processComponentAttachedListeners()` contains the same pattern:

```java
for (int i = 0, size = listeners.size(); i < size; i++)
{
    SystemEventListener listener = listeners.get(i);
    ...
}
```

If `holder.getListenersForEventClass()` ever returns a `CopyOnWriteArrayList` (which it commonly does in JSF implementations), this method has the identical race.

`processViewAttachedListeners()` is already written defensively by making a copy because of MYFACES-2935, so the author was clearly aware of concurrent modification there but overlooked the other two methods.

---

## I would also fix `subscribeToEvent()`

There is another concurrency issue:

```java
List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
if (eventInfos == null)
{
    eventInfos = new CopyOnWriteArrayList<>();
    globalListeners.put(systemEventClass, eventInfos);
}
```

Two threads can both create different lists:

```
T1: get() -> null
T2: get() -> null

T1: put(listA)
T2: put(listB)

T1: listA.add(listener1)
T2: listB.add(listener2)
```

One listener is effectively lost because `listA` is no longer referenced by the map.

This should instead use:

```java
List<EventInfo> eventInfos =
    globalListeners.computeIfAbsent(systemEventClass,
        k -> new CopyOnWriteArrayList<>());
```

which atomically creates or retrieves the shared list.

So I think there are **two real concurrency bugs** in `ApplicationImplEventManager`:

1. **`processGlobalListeners()`** (the one causing your exception) iterates a `CopyOnWriteArrayList` incorrectly using `size()`/`get()`.
2. **`subscribeToEvent()`** can lose listener registrations under concurrent subscriptions because it uses `get()`/`put()` instead of `computeIfAbsent()`.

Both are worth fixing in a MyFaces pull request.

tandraschko:
then you can still use:

List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
if (eventInfos == null)
{
    eventInfos = globalListeners.computeIfAbsent(systemEventClass, k -> new CopyOnWriteArrayList<>());
}

melloware:
pushed up some fixes

impl/src/main/java/org/apache/myfaces/application/ApplicationImplEventManager.java
```diff
@@ -122,11 +122,11 @@ public void subscribeToEvent(Class<? extends SystemEvent> systemEventClass, Clas
         Assert.notNull(systemEventClass, "systemEventClass");
         Assert.notNull(listener, "listener");
 
-        List<EventInfo> eventInfos = globalListeners.get(systemEventClass);
+        // MYFACES-4757 use computeIfAbsent to prevent multiple threads from creating different lists
+        List eventInfos = globalListeners.get(systemEventClass);
```
tandraschko: please readd generics

melloware:
Fixed

tandraschko:
+1 now

PR #1034, opened 2026-07-29
Key UIData/UIRepeat per-row EVH state by clientId, not row index

BalusC:
### Problem

A `UIData` or `UIRepeat` nested inside another iterating component loses the submitted values of every enclosing row but the last — silently: no conversion, no validation, no model update, no message.

Reproduced on a 5×10 grid (outer table over 5 groups, inner over 10 rows, 3 text inputs per row) by submitting a marker into every input and counting how many the response echoes back:

| scenario | before | after |
|---|--:|--:|
| `h:dataTable` in `h:dataTable` | 30/150 | **150/150** |
| `ui:repeat` in `ui:repeat` | 30/150 | **150/150** |
| composite in `h:dataTable` | 30/150 | **150/150** |
| flat table (control) | 105/105 | 105/105 |

The 30 that survived were exactly the last outer row.

### Cause

Regression from 6336caf1d ([perf] optimize UIData/UIRepeat state saving), which changed the per-row `EditableValueHolder` state map from `Map<String, …>` keyed by `getContainerClientId(context)` to `Map<Integer, …>` keyed by the bare row index.

A nested iterating component is one instance reused across the enclosing rows, so the bare index makes every enclosing row map to the same key. `getContainerClientId()` is the clientId plus the current row index, carrying both dimensions; the index key kept only one. `_rowDeltaStates` and `_rowTransientStates` were never changed and still key on the clientId.

### Fix

Key `_rowStates` on `getContainerClientId()` again in both components. The precollected iteration lists, the positional `EditableValueHolderState` list and its deferred allocation are untouched — only the key type and the two key expressions change.

**The optimisation is not being asked back.** Measured against current `main` with only `UIData`/`UIRepeat` reverted to their state before 6336caf1d, on the Mojarra `test/perf` suite (Tomcat, 32 scenarios, two alternating rounds):

| | pre-rewrite | this PR | Δ |
|---|--:|--:|--:|
| whole suite | 78.23 s | 76.36 s | **−2.4%** |
| APPLY_REQUEST_VALUES | 3.43 s | 3.09 s | **−10.0%** |
| UPDATE_MODEL_VALUES | 4.79 s | 4.41 s | **−7.9%** |

### Tests

`UIDataTest.testNestedTableRowStateIsScopedPerOuterRow` and `UIRepeatTest.testNestedRepeatRowStateIsScopedPerOuterRow`: a 2×2 nested grid, a distinct value per cell, iterate away and back, assert each cell kept its own. Both fail without the fix (`expected: <v00> but was: <v10>`) and pass with it. Full `api` + `impl` suite green, 1651 tests.

### Notes

- **A backport to `4.1.x` is needed**: b4a4050f6 carries the same change there, so 4.1.x has the same defect. This PR targets `main` only.
- Neither line has released it — 4.1.3 predates 6336caf1d, so no MyFaces release is affected. 4.0.x is not affected either (still clientId-keyed).
- Mojarra had the identical defect, fixed the same way in eclipse-ee4j/mojarra#5874.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 5)

### DIFF
--- a/api/src/main/java/jakarta/faces/component/UIData.java
+++ b/api/src/main/java/jakarta/faces/component/UIData.java
@@ -166,13 +166,13 @@ public class UIData extends UIComponentBase implements NamingContainer, UniqueId
     private int _rowIndex = -1;
 
     /**
-     * Per-row EVH state.  Outer key is the row index; value is the list of per-EVH states
-     * indexed by position in {@link #_iterationEVHList} (null entry = component in default
-     * empty state).  A missing outer entry also means all EVH components are in default state.
-     * Using an Integer row-index key avoids string concatenation and leverages the JVM Integer
-     * cache for typical row counts, making map operations cheaper than string-key alternatives.
+     * Per-row EVH state.  Outer key is {@link #getContainerClientId(FacesContext)}, i.e. this table's clientId plus
+     * its current row index, which for a table nested inside another iterating component also carries the enclosing
+     * rows' indices; value is the list of per-EVH states indexed by position in {@link #_iterationEVHList} (null
+     * entry = component in default empty state).  A missing outer entry also means all EVH components are in default
+     * state.
      */
-    private Map<Integer, List<EditableValueHolderState>> _rowStates = new HashMap<>();
+    private Map<String, List<EditableValueHolderState>> _rowStates = new HashMap<>();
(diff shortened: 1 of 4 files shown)


melloware (approved):

tandraschko:
Thanks!

tandraschko:
ported it to 4.1

PR #1043, opened 2026-08-13
[4.1.x] MYFACES-4761: Facelet path validation

volosied:


melloware (approved):

impl/src/main/java/org/apache/myfaces/view/facelets/impl/DefaultFaceletFactory.java
```diff
@@ -248,34 +256,205 @@ public long getRefreshPeriod()
...
+                    "Invalid path provided: " + path);
+        }
+
+        return resolved;
+    }
+
+    // Path-validation helpers
+
+    private static final Set<String> BLOCKED_SCHEMES = new HashSet<>(
+            Arrays.asList("http", "https", "ftp", "ftps", "mailto", "tel",
```
pnicolucci: Would it be better to have an allowed scheme(s) rather than blocked?

tandraschko:
Please provide some unittests

impl/src/main/java/org/apache/myfaces/view/facelets/impl/DefaultFaceletFactory.java
volosied: Fixed!

PR #1044, opened 2026-08-13
[5.0.x] MYFACES-4761: Facelet path validation

volosied:
TODO -- Description will be updated later.

### DIFF
--- a/impl/src/main/java/org/apache/myfaces/context/InvalidFileException.java
+++ b/impl/src/main/java/org/apache/myfaces/context/InvalidFileException.java
@@ -20,59 +20,27 @@ package org.apache.myfaces.context;
 
 import java.io.IOException;
 
-/**
- * Thrown when a Facelet resource path is rejected by the security or mapping
- * validation checks in the Facelet factory.
- * <p>
- * The {@link Reason} enum identifies which specific check triggered the rejection,
- * allowing callers to distinguish between a blocked URI scheme, a path that escapes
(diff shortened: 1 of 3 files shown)


tandraschko:
Im not sure If we should do this at this amount of detail.
In PrimeFaces we had a similar request If a user directly bind e.g. a request param via EL to a ImageCropper source. PF uses a URLConncection and would call this.
Its a bit a decision: should a framework catch every possible user error for 100%?
Maybe its enough to block external URL calls. If its path traversal, what could happen? MyFaces trying to parse anything else and throwing a exception, which leads to a Error Page?

melloware (approved):

volosied:
Hi, I removed some of the comments and simplified the implementation where possible.

However, the fix ended up being more involved because there are several edge cases where a simpler approach would break existing behavior and tests.

The main goal was to ensure that only valid Facelet resources within the application can be processed, while preventing access to other resources such as XML configuration files or arbitrary files. Simply blocking absolute paths is not sufficient, since behavior depends on how the resource is resolved, its location, and the URL scheme used by the container.

On the surface this looks like a straightforward issue, but in practice there are a number of legitimate resource resolution scenarios that need to continue working. The implementation tries to strike a balance between addressing the security concern and preserving existing functionality.

I will run TCK testing once more time, but this should be good unless you have any other concerns.

volosied:
Checks pass and unit tests pass. Merging. I'll work on other branches now.

PR #1046, opened 2026-08-17
[3.0.x] MYFACES-4761: Facelet path validation

volosied:


tandraschko:
Please also create a issue for Tracking 
And dont forget 4.0

PR #1066, opened 2026-08-26
4.0.X Fix CI worklow and remove Quarkus Extension from build

melloware:


melloware:
@tandraschko @volosied  this removes Quarkus and build is passing on 4.0.x again!

PR #1068, opened 2026-08-27
[4.1.x] Revert ChildSearchKeywordResolver.java to fix MYFACES-4764

volosied:
https://issues.apache.org/jira/browse/MYFACES-4764


tandraschko:
+1

please add a comment and test, otherwise it will be changed again

and also apply it to 5.0 then

melloware (approved):

PR #1075, opened 2026-09-01
Revert "[perf] use precompiled pattern"

volosied:
This reverts commit 5e56abd09afc9f1fbf08cf07b63dd75645866a19.

Binary compatibility tool reported that `jakarta.faces.validator.BeanValidator` changed:
image

 I cannot recall if this is allowed (new field), but I think the specs needs to match 1:1.


tandraschko:
Why not just make it private and duplicate it in the other class?

tandraschko:
But propably not worth, im ok with revert
Please also port it master

volosied:
I thought about that, but then I would have to declare the variable in three different classes and make it private.  

So I opted to revert it for simplicity.

melloware (approved):

PR #1084, opened 2026-09-22
MYFACES-4766:  BeanPropertyTagRule: stop caching arguments and use volatile to prevent concurrency errors

cocorossello:


tandraschko:
Why not just make the old field Lazy<>?

cocorossello:
> Why not just make the old field Lazy<>?

It's an option, but isn't this more complicated?

tandraschko:
Volatile/synchronized should not be required If you use it correctly. Lazy has a method arg, to resolve the value once

cocorossello:
Changed, but note that if getValue is not synchronized it may be initialized in more than 1 thread (it may be acceptable or better than synchronized anyway, I can remove it if you prefer)

tandraschko:
Lazy internally is synchronized
Please check it out

tandraschko:
Also we should avoid to instantiate the new array all the time

tandraschko:
Looks good for me, thanks
Can you test this in your environment, if its still fixed?

tandraschko:
@melloware build hangs, can you take over?

cocorossello:
Tested in my environment, no more cases, I applied last change today. To be absolutely sure it should be running a few days, because it only happened once or twice per day.

